PR #2, opened 2016-10-18
Adds more monoids

lloydmeta:
Adds `Sum` and `Product` monoids, while leaving the default numeric monoids to be
`Sum`.  

Example usage:

```rust
    #[test]
    fn test_combine_all_sum() {
        let v = vec![Sum(2), Sum(3), Sum(4)];
        assert_eq!(combine_all(&v), Sum(9))
    }

    #[test]
    fn test_combine_all_product() {
        let v = vec![Product(2), Product(3), Product(4)];
        assert_eq!(combine_all(&v), Product(24))
    }
```

@astynax I'm not sure if there's an issue with providing a default monoid to work on naked numbers, but my reasons for doing so are:

1. Efficiency (no need to wrap and unwrap at runtime)
2. Ergonomics
3. Scala libs like Cats provide Sum Monoids by default so at least it isn't unexpected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,9 @@ let v = vec![Some(1), Some(3)];
 assert_eq!(combine_all(&v), Some(4));
 
 // Slightly more magical
-let t1 = (1, 2.5f32, String::from("hi"), Some(3));
-let t2 = (1, 2.5f32, String::from(" world"), None);
-let t3 = (1, 2.5f32, String::from(", goodbye"), Some(10));
+let t1 =       (1, 2.5f32,                String::from("hi"),  Some(3));
+let t2 =       (1, 2.5f32,            String::from(" world"),     None);
+let t3 =       (1, 2.5f32,         String::from(", goodbye"), Some(10));
 let tuples = vec![t1, t2, t3];
 
 let expected = (3, 7.5f32, String::from("hi world, goodbye"), Some(13));
--- a/src/monoid.rs
+++ b/src/monoid.rs
@@ -208,7 +208,7 @@ tuple_impls! {
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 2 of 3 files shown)


astynax:
A a Haskell programmer, I can say what Sum and Product are pretty rarely present in real code. In the other hand threre are some uncommon but quite useful Monoids and/or Semigroups: bitwise `and`/`or`, `minimum`/`maximum` (these two - Semigroups but not Monoids).

However we can have a Monoid/Semigroups for naked numbers, which works as Sum, and put the `Product` into the library. In the same manner we can have a Monoid/Semigroup for the Booleans which work like `Or` (+ `And` wrapper in the library).

lloydmeta:
@astynax That sounds good. I've gone and added Max and Min monoids as well as`All` and `Any` for bitwise boolean logic.

Usage is as follows:

```rust

// Max
assert_eq!(
  combine_all_option(&vec![Max(1), Max(2), Max(3)]), 
  Some(Max(3))
);

// Min
assert_eq!(
  combine_all_option(&vec![Min(1), Min(2), Min(3)]), 
  Some(Min(1))
);

// Bitwise All (&&)
let v1: Vec<All<i32>> = Vec::new();
assert_eq!(combine_all(&v1), All(!0));
assert_eq!(combine_all(&vec![All(3), All(7)]), All(3));

// Bitwise Any (||)
let v2: Vec<Any<i32>> = Vec::new();
assert_eq!(combine_all(&v2), Any(0));
assert_eq!(combine_all(&vec![Any(3), Any(8)]), Any(11));


```

astynax:
Looks good! :+1:

lloydmeta:
Thanks for your time and feedback @astynax  !

PR #21, opened 2016-11-11
Try fixing Travis

lloydmeta:


.travis.yml
```diff
@@ -20,12 +20,12 @@ script:
 after_success:
   - travis-cargo doc-upload
   - if [ "${TRAVIS_PULL_REQUEST_BRANCH:-$TRAVIS_BRANCH}" != "master" ] && [ "$TRAVIS_RUST_VERSION" == "nightly" ]; then
```
QuLogic: You want to use a pipe to create a multi-line entry:
```
  - |
    if [ "....
    ...
    fi
```

lloydmeta: You sir, are a genious.

The `&&` i wanted to keep so that subsequent steps don't run if earlier ones fail.

lloydmeta: Ugh..now Travis can't checkout master. Any ideas?

QuLogic: I think because it clones a branch directly, you can't see any other refs. `git fetch origin master && git checkout -b master FETCH_HEAD` might work, or even `git fetch origin master:master`.

PR #24, opened 2016-11-18
Try git fetch origin master && git checkout -b master FETCH_HEAD

lloydmeta:


lloydmeta:
Hmm, checking out a new branch with the name of master with `FETCH_HEAD` just gives us a new branch with tip equal to the commit of the current branch.

QuLogic:
Ah, you've got two builds because this is also a branch on your repo. [This build](https://travis-ci.org/lloydmeta/frunk/jobs/176951210) is the nightly build for the _push_ (a branch on your repo) and [this build](https://travis-ci.org/lloydmeta/frunk/jobs/176951201) is the nightly build for the _PR_ (the merge of this PR with its base branch (master)).

The push build requires the `fetch && checkout -b` because Travis [clones the branch only](https://travis-ci.org/lloydmeta/frunk/jobs/176951210#L118). The PR build [clones this whole repo and then fetches and checks out the PR merge commit](https://travis-ci.org/lloydmeta/frunk/jobs/176951201#L118-L136), so in that case, you _don't_ need the `fetch && checkout -b` and should be able to checkout master directly.

lloydmeta:
I see. I think you're right :) That said, just cloning again into a different directory is just simpler..don't want to deal with conditionals if I don't need to :p 

Was fun figuring out what the problem was though !

PR #28, opened 2017-02-04
Allow identically structured HCons with combinable subtypes to be

tupshin:
combinable.

I was orignally trying to provide a concrete implementation of combine()
for all abstract HLists, but that was far less feasible.

This trivial implementation seems practically useful.


lloydmeta:
> I was orignally trying to provide a concrete implementation of combine()
for all abstract HLists, but that was far less feasible.

Yeah, I've run into this issue myself, which is unfortunate. Not entirely sure how it can be helped.

That said, this looks great 👍 Thanks !

PR #56, opened 2017-06-03
derive Hash

### LINKED ISSUE #55: derive Hash for HNil, HCons?
Is there a reason for these to not derive Hash? Otherwise you can't use HLists as keys in HashMap/HashSet.

mglagla:
Added `#[derive(Hash)]` to all types that are deriving anyway.

Closes #55


lloydmeta:
This is perfect. Thanks !

PR #74, opened 2017-10-06
Adds LiftFrom + LiftInto traits, Generic & From/Into for flat tuples <-> HList

Centril:
+ Bumps version numbers.
+ ignores `.vscode/`
+ `Here` becomes a struct instead - `Here/There` is just a normal type level `Nat` construct now.
+ Adds conversion traits `LiftFrom` and `LiftInto` allowing the users to construct HLists from an element or a sublist while defaulting everything not given.
+  Makes flat tuples up to `(T0, T1, T2, T3, T4, T5, T6, T7, T8, T9)` impl `Generic` as well as providing `From` + `Into` impls.

### DIFF
--- a/core/src/hlist.rs
+++ b/core/src/hlist.rs
@@ -1416,7 +1416,7 @@ mod tests {
     }
 
     #[test]
-    fn test_partial() {
+    fn test_lift() {
         type H = Hlist![(), usize, f64, (), bool];
 
         // Ensure type inference works as expected first:
--- a/core/src/hlist.rs
+++ b/core/src/hlist.rs
@@ -358,7 +358,7 @@ where
 ///
 /// Users should normally allow type inference to create this type
 #[allow(dead_code)]
-pub enum Here {}
+pub struct Here;
 
 /// Used as an index into an `HList`.
 ///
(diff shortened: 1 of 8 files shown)


lloydmeta:
Whoa! Thanks for this.  I'm out of town at the moment, but at a glance, it all looks super to me.  Will definitely give it a closer look as soon as I get back.

Centril:
Wonderful! I might have one or two more commits to add with more `From` impls etc. to make the usage of HLists even more ergonomic. If so, I'll add them today.

core/src/lib.rs
```diff
@@ -61,6 +61,7 @@
 pub mod hlist;
 pub mod generic;
 pub mod labelled;
+mod tuples;
```
lloydmeta: Is it ok for this to not be `pub` ? I'm not sure if the implementations are useable outside this crate as is.

core/src/hlist.rs
```diff
@@ -358,7 +358,7 @@ where
 ///
 /// Users should normally allow type inference to create this type
 #[allow(dead_code)]
-pub enum Here {}
+pub struct Here;
```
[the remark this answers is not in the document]
Centril: Not really, it doesn't have to be a struct. Tho it makes more sense mathematically ;)

core/src/lib.rs
Centril: Iirc, implementations belong to the crate and the trait belongs to a module. So it should be fine...
You can verify this by checking `cargo doc --open`.

core/src/hlist.rs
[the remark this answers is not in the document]
lloydmeta: Ah, in that case let's have this back as a zero-member enum; it served the practical purpose of signaling that `Here` is a compile-time construct used for type-level purposes and has zero runtime cost, since it can't be instantiated.

core/src/lib.rs
lloydmeta: Oh, interesting. Didn't know that! Thanks, verified it locally. Should be fine.

lloydmeta:
I'll merge this now and make modifications later. Thanks again!

PR #76, opened 2018-01-12
Fix documentation on monoid.rs

KiChjang:
Adds a comment describing the contents of h3, which makes it consistent with the other HashMaps.


lloydmeta:
Ah nice. Thanks !

PR #81, opened 2018-02-22
AUnion

remexre:
This add anonymous extensible unions, which might have a "real" name I don't know.

Essentially, they're a type with the operations:

```haskell
class AUnion t u a where
  inject :: t -> a
  select :: a -> Either t u
```

I (or rather, Vim) also ran rustfmt. Also, fixed a typo in the README.

---

Things I should do:

 - [x] Write docs in a long-form comment
 - [x] Write unit tests
 - [x] Add to examples
 - [x] Replace `Nat` with `Here`/`There` (should I actually do this?)


remexre:
Ping @Centril, I know you were interested in this.

Centril:
Some thoughts of mine:
1. You don't need to introduce `AUnionNil` and `AUnionCons`, use [`CNil`](https://docs.rs/frunk_core/0.0.22/frunk_core/coproduct/enum.CNil.html) and [`Coproduct`](https://docs.rs/frunk_core/0.0.22/frunk_core/coproduct/enum.Coproduct.html) instead.
2. We already have [`CoprodInjector`](https://docs.rs/frunk_core/0.0.22/frunk_core/coproduct/trait.CoprodInjector.html) for the `inject` part. Does that not cover that specific trait function?
3. What you are proposing is a more expressive form of [`CoproductTaker`](https://docs.rs/frunk_core/0.0.22/frunk_core/coproduct/trait.CoproductTaker.html) so I think we should use a similar naming.
4. What about by shared reference as in [`CoproductSelector`](https://docs.rs/frunk_core/0.0.22/frunk_core/coproduct/trait.CoproductSelector.html) and exclusive (mutable) reference?
5. I think the additions should be part of the `coproduct` module and in `frunk_core`.

remexre:
Fixed 1 and 2.

> What you are proposing is a more expressive form of CoproductTaker so I think we should use a similar naming.
How does `CoprodUninject` sound?

> What about by shared reference as in CoproductSelector and exclusive (mutable) reference?
`uninject`, `uninject_ref`, `uninject_mut`?

> I think the additions should be part of the coproduct module and in frunk_core.
Sure, I'll move it when I change the name.

Centril:
cc @lloydmeta ^

I think those work. It might be worth refactoring when GATs land, but we shouldn't await that.

remexre:
Oh, remembered/realized why the ref/mut methods don't work -- the `CoProdUninjector<T, Coproduct<Hd, U>, There<N>>` case needs me to write a function:

```rust
fn uninject_ref(&self) -> Result<&T, &Coproduct<Hd, U>> { ... }
```

and that doesn't work :/

Centril:
Is that perhaps fixable with an associated type that pushes the `&` inwards in the tail, so you get `Result<&T, Self::Tail>` and then `Tail` becomes `-> Result<&T, Coproduct<&Hd, ...>>`?

remexre:
I don't think so -- the method is then:

```rust
    fn uninject_ref(&'a self) -> Result<&'a T, Self::RefTail> {
        match *self {
            Coproduct::Inl(ref h) => Err(Coproduct::Inl(&h)),
            Coproduct::Inr(ref t) => t.uninject_ref().map_err(Coproduct::Inr),
            //                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
            // type error: This is Err(Tl::RefTail), not Err(Self::RefTail).
        }
    }
```

remexre:
Would it make sense to make `as_ref : &CoProdUninjector!(T, U, V) -> CoProdUninjector!(&T, &U, &V)` and  `as_mut : &mut CoProdUninjector!(T, U, V) -> CoProdUninjector!(&mut T, &mut U, &mut V)`? That might be more possible; I'll try it later tonight.

lloydmeta:
Thanks @remexre for the cool idea and @Centril for reviewing + guidance.

I really like it so far, and if you cant get the ref variation going, I agree with @Centril: let's get the `uninject` in (with tests) and not let perfect be the enemy of good.

remexre:
The ref version seems hard (or at least harder than I have time for in the short term). I've got all the tasks in the original post done; is this good to go, or should I squash it, or does it need more work?

Centril:
LGTM 👍

lloydmeta:
Awesome stuff :) 

Thanks for the awesome work @remexre; you went above and beyond by fixing the formatting *and* my (seriously embarrassing 😅  ) naming bloopers.

Thanks again @Centril for doing reviews and guidance (are you open to becoming a maintainer ? :D)

Centril:
> Thanks again @Centril for doing reviews and guidance (are you open to becoming a maintainer ? :D)

Sure, I'm open to that =)

PR #88, opened 2018-03-19
Change CoprodUninjector U parameter to an associated type Remainder

### LINKED ISSUE #87: Why is CoprodUninjector's Remainder a type parameter and not an associated type?
When I recently wrote my own coproduct type I had no difficulty defining the remainder of type-indexed removal from a coproduct [as an associated type](https://github.com/ExpHP/funky/blob/e475f7844e124bb7e02c2b38967e2b6aa5e373d1/src/coprod.rs#L421-L443).  ISTM this is vastly preferrable to an "always-inferred type parameter," which feels flimsy, is difficult to name and will inevitably show up uninvited in other function's signatures.

cc @remexre, do you recall the reason for adding `U` as a type parameter?

ExpHP:
Closes #87

I also snuck in a fix for a typo in one of the unit tests.

### DIFF
--- a/core/src/coproduct.rs
+++ b/core/src/coproduct.rs
@@ -373,12 +373,16 @@ impl<CH, CTail> AsRef<Coproduct<CH, CTail>> for Coproduct<CH, CTail> {
 /// assert_eq!(get_from_1b, Ok(42f32));
 /// # }
 /// ```
-pub trait CoprodUninjector<T, U, Idx>: CoprodInjector<T, Idx> {
+pub trait CoprodUninjector<T, Idx>: CoprodInjector<T, Idx> {
+    type Remainder;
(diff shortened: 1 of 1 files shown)


core/src/coproduct.rs
```diff
@@ -496,7 +502,7 @@ mod tests {
         let uninject_i32_co3: Result<i32, _> = co3.uninject();
         let uninject_str_co3: Result<&'static str, _> = co3.uninject();
         let uninject_bool_co3: Result<bool, _> = co3.uninject();
-        assert!(uninject_i32_co2.is_err());
+        assert!(uninject_i32_co3.is_err());
```
lloydmeta: Wow nice catch

Issue #86, opened 2018-03-19
Splitting coproducts, and upcasting to larger coproducts


ExpHP:
Over the weekend I've been working on a [reimplementation of frunk](https://github.com/ExpHP/funky/tree/master) to help understand the current design and gather ideas.  The following two ideas are *really low hanging fruit;* I mean, when I wrote them, I just figured that frunk _obviously_ must already have something like these (but looking around, I don't think it does).

I can add something similar to frunk if others would like them.

---

#### Splitting coproducts

Split e.g. `Coprod![A, B, C, D, E]` into `Result<Coprod![D, A, B], Coprod![C, E]>`.

(In my repo I named it `sculpt` after the similar operation on HLists, but I'm not too sure about that name)

```rust
// Indices is usually inferred
pub trait CoprodSculptor<Targets, Indices>: Sized {
    type Remainder;

    fn sculpt(self) -> Result<Targets, Self::Remainder>;
}
```

The [implementation](https://github.com/ExpHP/funky/blob/e475f7844e124bb7e02c2b38967e2b6aa5e373d1/src/coprod.rs#L528-L565) is just repeated calls to `uninject()`. (`.pluck()` in my codebase)

#### Upcasting

Upcast e.g. `Coprod![A, B, D]` into `Coprod![A, B, C, D, E]`.

```rust
// Indices is usually inferred
pub trait CoprodEmbedder<Choices, Indices> {
    fn embed(self) -> Choices;
}
```

My [implementation of `embed`](https://github.com/ExpHP/funky/blob/e475f7844e124bb7e02c2b38967e2b6aa5e373d1/src/coprod.rs#L633-L654) is insultingly simple, and if rust had generic closures it could be written as just `coprod.fold(|x| Target::inject(x))`.  Some neat properties that fall out of this simple implementation are that order doesn't matter, and duplicate types in the input can map to the same place in the output (so `Coprod![C, A, E]` and `Coprod![B, B, B, B]` can also be converted into `Coprod![A, B, C, D, E]`).

---

[...]

PR #89, opened 2018-03-20
add CoprodSubsetter for splitting coproducts

ExpHP:
Part of #86.

* Adds `CoprodSubsetter` with a `subset` method to split a coproduct.  The idea behind the name is that you are handling a subset of the types, leaving the rest for somebody else to handle.

* Added an example of using CoprodInjector for a match that is actually exhaustive.  It didn't seem right for only `CoprodSubsetter` to have such an example when `CoprodInjector` has the word "exhaustive" right there in the headline.

### DIFF
--- a/core/src/coproduct.rs
+++ b/core/src/coproduct.rs
@@ -385,7 +385,7 @@ impl<CH, CTail> AsRef<Coproduct<CH, CTail>> for Coproduct<CH, CTail> {
 /// # fn main() {
 /// type I32F32 = Coprod!(i32, f32);
 ///
-/// // be aware that this particular example could be
+/// // Be aware that this particular example could be
 /// // written far more succinctly using `fold`.
 /// fn handle_i32_f32(co: I32F32) -> f32 {
 ///     // Remove i32 from the coproduct
@@ -402,7 +402,7 @@ impl<CH, CTail> AsRef<Coproduct<CH, CTail>> for Coproduct<CH, CTail> {
 ///         Err(co) => co
 ///     };
 ///
-///     // now co is empty;
+///     // Now co is empty;
 ///     match co { /* unreachable */ }
 /// }
 ///
@@ -499,13 +499,13 @@ where
 ///         Err(co) => co,
 ///     };
 ///
-///     // now co is Coprod!(u32)
+///     // Now co is Coprod!(u32).
 ///     let co = match co.subset().map(handle_countly_things) {
 ///         Ok(s) => return s,
 ///         Err(co) => co,
 ///     };
 ///
-///     // now co is empty;
+///     // Now co is empty.
 ///     match co { /* unreachable */ }
 /// }
 ///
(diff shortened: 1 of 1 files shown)


core/src/coproduct.rs
```diff
@@ -359,20 +359,56 @@ impl<CH, CTail> AsRef<Coproduct<CH, CTail>> for Coproduct<CH, CTail> {
...
+///
+/// Chaining calls for an exhaustive match:
+///
+/// ```rust
+/// # #[macro_use] extern crate frunk_core;
+/// # use frunk_core::coproduct::*;
+/// # fn main() {
+/// type I32F32 = Coprod!(i32, f32);
+///
+/// // be aware that this particular example could be
```
Centril: s/be/Be

core/src/coproduct.rs
```diff
@@ -405,6 +441,112 @@ where
...
+///     // now co is empty;
+///     match co { /* unreachable */ }
+/// }
+///
+/// assert_eq!(handle_anything(Anything::inject("hello")), "&str hello");
+/// assert_eq!(handle_anything(Anything::inject(String::from("World!"))), "String World!");
+/// assert_eq!(handle_anything(Anything::inject(4)), "....");
+/// # }
+/// ```
+pub trait CoproductSubsetter<Targets, Indices>: Sized {
```
Centril: I wonder if this needs a separate trait or if it would be possible to simply do this as impls of `CoproductUninjector` with a label type?

Centril:
LGTM in general 👍

core/src/coproduct.rs
ExpHP: What do you mean by label type? In general I am fearful of having these be the same method (especially if it is entirely driven by type inference) because of the ambiguities created when you have a coproduct containing coproducts.

Centril: Take a look at:
+ https://docs.rs/frunk_core/0.0.23/frunk_core/hlist/trait.LiftInto.html
+ https://docs.rs/frunk_core/0.0.23/frunk_core/hlist/trait.LiftFrom.html
+ https://docs.rs/frunk_core/0.0.23/frunk_core/hlist/struct.Suffixed.html -- here's the index type.

Inference problems is a good point 👍

ExpHP: @Centril That is the most criminally underdocumented and bizarre feature I have ever seen.

(Well, second most bizarre.  The *most* bizarre feature I've seen are the impls of `From` and `Into` for tuples of the wrong size, sitting there right next to the impls for tuples of matching size)

* The documentation of `lift_from` shows HLists being produced from single values.
* Why on earth is this same function also used to generate an HList from a prefix?  Why a prefix and not a suffix?  Why not an arbitrarily ordered subset?
* It shares my aforementioned concern about ambiguous input types.

Centril: > That is the most criminally underdocumented and bizarre feature I have ever seen.

Feel free to improve the documentation. The docs for `LiftInto` shows Hlists being produced from multiple values, but its out of sync with `LiftFrom`.

> Why a prefix and not a suffix? Why not an arbitrarily ordered subset?

Prefixes are supported afaik.
Arbitrarily ordered subsets never crossed my mind; If you can improve on that, it would be nice.

Issue #94, opened 2018-03-21
Documenting rules for when type inference can solve for indices


ExpHP:
I do a small bit of this in the coproduct `embed` PR, but I think there should be more:

There should be one central piece of documentation (which other methods can link to) describing the `Index` technique commonly used in frunk, describing how:

* Users should generally not worry about `Index` type parameters.  Even when using turbofish syntax (once available), these should generally be left to type inference using `_`.
* Generally speaking, type inference can solve for the indices whenever there is a **unique solution.**
* This means that methods like hlist's `pluck` and coproduct's `inject` succeed **when the type appears exactly once** in the HList/coproduct.

It should also explain workarounds for when type inference fails:

* **The preferred method, whenever possible, should be to use newtypes in your HList/coproduct,** so that all of the types are distinct. (@Centril, @lloydmeta do you agree?)
* As a last resort, indices can be specified explicitly, but I'm not sure that the docs should be suggesting this course of action quite yet. (maybe once reified indices are added, if I ever get to that; I still need to get more experience with them first in [`funky`](https://github.com/ExpHP/funky))

Centril:
> Users should generally not worry about `Index` type parameters. Even when using turbofish syntax (once available), these should generally be left to type inference using `_`.

Agreed; Hopefully you won't even need to specify `_` in the future with some sort of "partial turbofish" (see my RFC on this -- https://github.com/rust-lang/rfcs/pull/2176) mechanism in the language itself.

I also agree with everything else you said.

PR #95, opened 2018-03-21
Feature/CoproductEmbedder

### LINKED ISSUE #86: Splitting coproducts, and upcasting to larger coproducts

ExpHP:
## Itemized changes

From most significant to least:

**Added `CoproductEmbedder`.**

**Documented the rules of type inference for `CoprodInjector`.** I did this because I allude to these rules in the docs for `CoproductEmbedder`. (In the future, I would like to see [similar documentation](https://github.com/lloydmeta/frunk/issues/94) added to all methods.)

**Changed headline of `CoprodInjector` to not use the word "inject."**

**Removed `#[doc(hidden)]` changes from some trait impls for `CNil`.** In my opinion it actually makes the documentation nicer when the reader can see the base cases for recursion.
*(@lloydmeta let me know if you disagree and I can undo this)*

**Add missing code tags to some markdown.**

---

Closes #86.

### DIFF
--- a/core/src/coproduct.rs
+++ b/core/src/coproduct.rs
@@ -142,7 +142,14 @@ macro_rules! Coprod {
     // Forward trailing comma variants -->
 }
 
-/// Trait for injecting something into a coproduct
+/// Trait for instantiating a coproduct
+///
+/// This is the preferred method for constructing a coproduct.
+/// In standard usage, the index type parameter can be ignored,
+/// as it will typically be solved for using type inference.
+///
+/// If the type does not appear in the coproduct, the conversion is forbidden.
+/// If the type appears multiple times in the coproduct, type inference will fail.
 ///
 /// ```
 /// # #[macro_use] extern crate frunk_core;
@@ -180,7 +187,7 @@ where
 
 /// Trait for retrieving a coproduct element reference by type.
 ///
-/// Returns an Option<&YourType> (notice that the inside of the option is a reference)
+/// Returns an `Option<&YourType>` (notice that the inside of the option is a reference)
 ///
 /// # Example
 ///
@@ -228,7 +235,7 @@ where
 
 /// Trait for retrieving a coproduct element by type.
 ///
-/// Returns an Option<YourType> (notice that the inside of the option is a value)
+/// Returns an `Option<YourType>` (notice that the inside of the option is a value)
 ///
 /// # Example
 ///
@@ -336,7 +343,6 @@ where
 }
 
 /// This is literally impossible; CNil is not instantiable
-#[doc(hidden)]
 impl<F, R> CoproductFoldable<F, R> for CNil {
     fn fold(self, _: F) -> R {
         unreachable!()
@@ -344,7 +350,6 @@ impl<F, R> CoproductFoldable<F, R> for CNil {
 }
 
 /// This is literally impossible; &CNil is not instantiable
-#[doc(hidden)]
 impl<'a, F, R> CoproductFoldable<F, R> for &'a CNil {
     fn fold(self, _: F) -> R {
         unreachable!()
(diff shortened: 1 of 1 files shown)


core/src/coproduct.rs
```diff
@@ -547,6 +552,75 @@ impl<Choices> CoproductSubsetter<CNil, HNil> for Choices {
     { Err(self) }
 }
 
+/// Trait for upcasting a coproduct into another that can hold its variants.
```
Centril: Hmm... is "upcasting" the right word here?

Centril:
Assigning over to @lloydmeta since I'm a bit busy atm ;)

My initial reaction is that this is a great idea 👍

core/src/coproduct.rs
ExpHP: It is a bit misleading, I agree.  (my original goal when first designing it was to make something that does upcasting; but the final implementation is capable of doing other things as well)

What I initially wrote here was "Trait for infallible conversions between coproducts", though I was afraid somebody might ask why not use `From/Into` for this.  (to which my reply would be: we only get one chance to write a generic from/into, so we have to spend it carefully)

Centril: Can we just say the following and let the examples explain the semantics?
> Trait for **embedding** a coproduct into another that can hold its variants.

ExpHP: I was thinking of how it would look on e.g. the module page (where that first line is the only information a user is given). How will a user know that "embedding" is what they want?

That said, this concern won't matter once inherent method wrappers are added (in which case the first line doesn't show up out of context anywhere, and so "embedding" would work just fine).

ExpHP: How about just

> Trait for **converting** a coproduct into another that can hold its variants.

Centril: SGTM 👍

PR #98, opened 2018-03-22
Follow conventional feature-gating for serde

### LINKED ISSUE #97: (C-SERDE) the serde feature should be named "serde"
See [C-SERDE](https://rust-lang-nursery.github.io/api-guidelines/interoperability.html#data-structures-implement-serdes-serialize-deserialize-c-serde) in the Rust API guidelines

Note it is not actually possible to name a feature the same as a dependency. Rather, [`serde` provides a `"derive"` feature](https://github.com/serde-rs/serde/blob/7ad836e6a97d7cc08cb36bc962745f3262208f66/serde/src/lib.rs#L238-L276) specifically for this purpose.

lloydmeta:
Closes #97

* Follow guidelines in https://rust-lang-nursery.github.io/api-guidelines/interoperability.html#data-structures-implement-serdes-serialize-deserialize-c-serde
* Add serde derivation for Coproducts as well

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,11 +28,7 @@ path = "laws"
 version = "0.0.14"
 
 [dependencies]
-serde = { version = "^1.0", optional = true, default-features = false }
-serde_derive = { version = "^1.0", optional = true, default-features = false }
-
-[features]
-with-serde = ["serde", "serde_derive"]
+serde = { version = "^1.0", optional = true, features = [ "derive" ] }
 
 [profile.bench]
 opt-level = 3
--- a/README.md
+++ b/README.md
@@ -36,10 +36,11 @@ For a deep dive, RustDocs are available for:
 4. [Validated](#validated)
 5. [Semigroup](#semigroup)
 6. [Monoid](#monoid)
-7. [Todo](#todo)
-8. [Contributing](#contributing)
-9. [Inspirations](#inspirations)
-10. [Maintainers](#maintainers)
+7. [Features](#features)
+8. [Todo](#todo)
+9. [Contributing](#contributing)
+10. [Inspirations](#inspirations)
+11. [Maintainers](#maintainers)
 
 ## Examples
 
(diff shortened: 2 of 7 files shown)


ExpHP:
`frunk` should also do this, not just core.

But hmmm... I dont think it will be possible to have the `serde` feature of frunk toggle the `serde` feature of core when using this hack. Let me think about this...

ExpHP:
Here's how `num` does it:

https://github.com/rust-num/num/blob/master/Cargo.toml

Basically, `num` itself doesn't contain anything, so it doesn't need serde as a dependency. That allows "serde" to be used as a feature name.

For `frunk` to be able to do that, everything in frunk that doesn't come from `frunk-core` would need to be split out into a third crate. That's kind of a big change! :P

Still looking for a better alternative...

ExpHP:
Okay, so it looks like having the features be disconnected from each other is already the status quo for frunk, so this is my recommendation for now:

* Make the same changes to frunk's Cargo.toml.
* Add a note to README.md that enabling serde requires one to explicitly depend on `frunk-core`: (note: this was already true before this PR, but I think it is worth documenting now)

```toml
[dependencies]
frunk = { version = "0.1.36", features = ["serde"] }
frunk-core = { version = "0.1.36", features = ["serde"] }
```

lloydmeta:
@ExpHP Aaah right; good point. I've made the same change to frunk and updated the readme. Let me know what you think.

ExpHP:
Hm, that's funny. I didn't realize this earlier, but: judging from how the second commit doesn't touch any `.rs` files, I guess there isn't actually anything in `frunk` itself that *has* serde derives? :stuck_out_tongue: 

Okay, so technically you *can* do what `num` does, and add a `serde` feature to `frunk` that activates the feature `"frunk-core/serde"`.  The advantage of this would be that a user doesn't need to depend on `frunk-core` explicitly.

**Or:** what you just commited is also fine (even if it seems a bit silly at the moment).  The advantage of this choice is that `frunk` will be able to backwards-compatibly add things that use serde in the future.

PR #99, opened 2018-03-23
Inherent methods for core types

ExpHP:
Okay, I think this is ready.

---

**Todo:**
* ~Thoroughly look over the documentation and:~
    * ~Make sure method order is nice~
    * ~Double check content of docstrings~
    * ~Double check all links~

---

<details>
<summary>Original text</summary>
This currently just covers HList because I wanted to know if there is any feedback before continuing with Coproduct.

The actual addition of the inherent methods is only a small part of this.  By far the biggest aspect of such a feature is documentation, because without good documentation, having both the trait methods and the inherent methods will create confusion.

* Docs on traits were moved to the new methods where appropriate.
* Examples in `hlist.rs` are adjusted to show turbofish
* Most trait docs have significantly changed.
  * `Selector`, `Sculptor`, `Plucker` all direct the reader's attention toward the inherent method, leaving the trait for just generic code.
  * IntoTuple2 says "you should never need to import this trait" because it is not useful in generic code.
  * IntoReverse was allowed to keep its headline.  (I can see other types implementing it in the future.)  It just casually remarks "This functionality is also provided as an inherent method."

</details>

### DIFF
--- a/core/src/coproduct.rs
+++ b/core/src/coproduct.rs
@@ -485,7 +485,7 @@ impl<Head, Tail> Coproduct<Head, Tail> {
 /// Coproducts of unknown type. In most code, `Coproduct::inject` will
 /// "just work," with or without this trait.
 ///
-/// [`Coproduct::inject`]: ../enum.Coproduct.html#method.inject
+/// [`Coproduct::inject`]: enum.Coproduct.html#method.inject
 pub trait CoprodInjector<InjectType, Index> {
     /// Instantiate a coproduct from an element.
     ///
@@ -495,7 +495,7 @@ pub trait CoprodInjector<InjectType, Index> {
     /// trait method is the location of the type parameters.
     /// (here, they are on the trait rather than the method)
     ///
-    /// [inherent static method]: ../enum.Coproduct.html#method.inject
+    /// [inherent static method]: enum.Coproduct.html#method.inject
     fn inject(to_insert: InjectType) -> Self;
 }
 
@@ -526,7 +526,7 @@ where
 /// Coproducts of unknown type. If you have a Coproduct of known type,
 /// then `co.get()` should "just work" even without the trait.
 ///
-/// [`Coproduct::get`]: ../enum.Coproduct.html#method.get
+/// [`Coproduct::get`]: enum.Coproduct.html#method.get
 pub trait CoproductSelector<S, I> {
     /// Borrow an element from a coproduct by type.
     ///
@@ -536,7 +536,7 @@ pub trait CoproductSelector<S, I> {
     /// trait method is the location of the type parameters.
     /// (here, they are on the trait rather than the method)
     ///
-    /// [inherent method]: ../enum.Coproduct.html#method.get
+    /// [inherent method]: enum.Coproduct.html#method.get
     fn get(&self) -> Option<&S>;
 }
 
@@ -573,7 +573,7 @@ where
 /// Coproducts of unknown type. If you have a Coproduct of known type,
 /// then `co.take()` should "just work" even without the trait.
 ///
-/// [`Coproduct::take`]: ../enum.Coproduct.html#method.take
+/// [`Coproduct::take`]: enum.Coproduct.html#method.take
 pub trait CoproductTaker<S, I> {
     /// Retrieve an element from a coproduct by type, ignoring all others.
     ///
@@ -583,7 +583,7 @@ pub trait CoproductTaker<S, I> {
     /// trait method is the location of the type parameters.
     /// (here, they are on the trait rather than the method)
     ///
-    /// [inherent method]: ../enum.Coproduct.html#method.take
+    /// [inherent method]: enum.Coproduct.html#method.take
     fn take(self) -> Option<S>;
 }
 
@@ -701,7 +701,7 @@ impl<CH, CTail> AsRef<Coproduct<CH, CTail>> for Coproduct<CH, CTail> {
 /// Coproducts of unknown type. If you have a Coproduct of known type,
 /// then `co.uninject()` should "just work" even without the trait.
 ///
-/// [`Coproduct::uninject`]: ../enum.Coproduct.html#method.uninject
+/// [`Coproduct::uninject`]: enum.Coproduct.html#method.uninject
 pub trait CoprodUninjector<T, Idx>: CoprodInjector<T, Idx> {
     type Remainder;
 
@@ -713,7 +713,7 @@ pub trait CoprodUninjector<T, Idx>: CoprodInjector<T, Idx> {
     /// trait method is the location of the type parameters.
     /// (here, they are on the trait rather than the method)
     ///
-    /// [inherent method]: ../enum.Coproduct.html#method.uninject
+    /// [inherent method]: enum.Coproduct.html#method.uninject
     fn uninject(self) -> Result<T, Self::Remainder>;
 }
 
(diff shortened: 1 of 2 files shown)


lloydmeta:
Looks great far so good to me; my one question ATM is with the moving of doc-tests for HNil and HCons; do these run fine?

It looks like they do on Travis, just double checking:
- This looks like the doc-test for `HCons`'s len: https://travis-ci.org/lloydmeta/frunk/jobs/357234016#L784
- This looks like the doc-test for `HNil`'s len: https://travis-ci.org/lloydmeta/frunk/jobs/357234016#L815

ExpHP:
They do run, although they are currently duplicates of each other.

Fun fact about doc comments: They get transformed into `#[doc = ... ]` annotations before  the token stream is given to macros, so we can even match them explicitly in the macro.  I plan to use this in the future to give different docstrings to some of the HNil functions.

lloydmeta:
> Fun fact about doc comments: They get transformed into `#[doc = ... ]` annotations before the token stream is given to macros, so we can even match them explicitly in the macro. I plan to use this in the future to give different docstrings to some of the HNil functions.

Interesting. TIL!

ExpHP:
@Centril I get the impression that you generally prefer having a few general interfaces over lots of little feature-specific interfaces, so I want to make sure you are okay with the reduced roles I am giving each trait.

(P.S. not all of these are done in the PR yet.   These are just my plans)

The following traits will be **unaffected by my PR** because I am not giving them inherent methods:

* `hlist::LiftFrom`, `hlist::LiftInto`
* `HMappable`, `HFoldLeftable`, `HFoldRightable`
* `CoproductFoldable`

The following traits still describe themselves as being **intended for general types.** (I.e. not necessarily specific to HLists or Coproducts)

* `hlist::IntoReverse`

The following traits describe themselves as **a bound for using a method of HList (or Coproduct) in generic code:**

* `hlist::Selector`
* `hlist::Plucker`
* `hlist::Sculptor`
* `CoprodInjector`
* `CoprodUninjector`
* `CoproductSelector`
* `CoproductTaker`
* `CoproductSubsetter`
* `CoproductEmbedder`

The following traits now describe themselves as **purely implementation detail** of a method on HList or Coproduct.  There is no good reason to ever need the trait in generic code.

* `hlist::IntoTuple2`

Centril:
It sounds good in general 👍 

A small thought...

> The following traits describe themselves as **a bound for using a method of HList (or Coproduct) in generic code:**

To increase the usefulness of frunk, we could also let some of these traits operate on other product types such as built in tuples. I agree that we should mainly consider / describe the traits as bounds for hlists, etc., but we could also consider them more generally? Tho the `Generic` trait allows you to go back and forth between tuples and hlists.

ExpHP:
IMHO, I'd prefer to have general functionality like that on a separate trait, one which is designed to be imported.  That trait can then be `impl`'d on `HCons` where `Self: hlist::Selector`, and use different bounds on other types where appropriate.  (or maybe can be impl'd on everything where `Self: Generic`).

This way, the design of `Selector` can remain optimized for the implementation `HList::get`, without potentially needing to change to accommodate similar methods on other types.

Centril:
Testing out `T: Generic<Repr = some_hlist>` sounds interesting; For now, I'm comfortable with your general plan. If we can add more impls on tuples that do not hurt type inference later, that could be nice but not strictly necessary.

ExpHP:
I was really afraid of how these signatures would look in documentation, so I'd just like to take a moment to thank rustdoc for being **goddamn awesome!**

```rust
pub fn subset<Targets, Indices>(self)    // v--- yikes!
-> Result<Targets, <Self as CoproductSubsetter<Targets, Indices>>::Remainder>
where
    Self: CoproductSubsetter<Targets, Indices>,
```

screenshot_20180324_131909

core/src/coproduct.rs
```diff
@@ -144,28 +144,360 @@ macro_rules! Coprod {
...
+    ///
+    /// // You can also use turbofish syntax to specify the type.
+    /// // The Index parameter should be left as `_`.
+    /// let co2 = I32F32::inject(1i32);
+    /// assert_eq!(co2.get::<i32, _>(), Some(&1));
+    /// assert_eq!(co2.get::<f32, _>(), None);
+    /// # }
+    /// ```
+    #[inline(always)]
+    pub fn get<S, Index>(&self) -> Option<&S>
```
Centril: A thought that occurred to me just now...
Should we generalize this to `-> R` where `R: Try<&S, E>` once `Try` is stable?
(This applies generally where we have `Option<?>`)

Centril:
We probably want to run some rustfmt on all the code later ;)

core/src/coproduct.rs
ExpHP: Eh, I'm not sure if I like it. Where does the `E` come from?  ISTM that E would have to `impl Default` or `impl From<NoneError>` or something like that and I'm not sure if there's an obvious "right choice"

ExpHP: Also it would make it difficult to do things like `self.get().expect()` or `self.get().map(f)`

Centril: I think `impl Default` would be reasonable, but I hear you on `self.get().map(f)` -- type inference is not helped.

ExpHP:
@lloydmeta I think this is ready :slightly_smiling_face:

core/src/coproduct.rs
```diff
@@ -144,28 +144,361 @@ macro_rules! Coprod {
     // Forward trailing comma variants -->
 }
 
-/// Trait for instantiating a coproduct
+// Inherent methods
+impl<Head, Tail> Coproduct<Head, Tail> {
+    /// Instantiate a coproduct from an element.
+    ///
+    /// This is generally much nicer than nested usage of `Coproduct::{Inl,Inr}`.
```
Centril: Space between `Inl,` and `Inr` ;)

core/src/coproduct.rs
```diff
@@ -144,28 +144,361 @@ macro_rules! Coprod {
...
+    /// It uses a trick with type inference to automatically build the correct variant
```
Centril: s/It/The function/

core/src/coproduct.rs
```diff
@@ -144,28 +144,361 @@ macro_rules! Coprod {
...
+    /// according to the input type.
+    ///
+    /// In standard usage, the Index type parameter can be ignored,
```
Centril: Use backticks on `Index`

core/src/coproduct.rs
ExpHP: I'll do this, but until now I've always *deliberately* written it without the space since the `{a,b}` syntax comes from shell syntax, where a space would actually mess things up. Might be a tough habit to unlearn... :stuck_out_tongue:

core/src/coproduct.rs
ExpHP: thanks

core/src/coproduct.rs
```diff
@@ -144,28 +144,361 @@ macro_rules! Coprod {
...
+    ///         Ok(x) => return 2.0 * x,
+    ///         Err(co) => co,
+    ///     };
+    ///
+    ///     // Now co is empty
+    ///     match co { /* unreachable */ }
+    /// }
+    ///
+    /// assert_eq!(handle_i32_f32(I32F32::inject(3)), 6.0);
+    /// assert_eq!(handle_i32_f32(I32F32::inject(3.0)), 6.0);
```
lloydmeta: Nit: I wonder if this test would be more illustrative if the results from `handle_i32_f32` were different when `i32` is pulled out vs `f32` (e.g. one multiplies by 2, the other by 3)

PR #102, opened 2018-03-25
add ...rest patterns to macros, and test structs

### LINKED ISSUE #100: Rest pattern in hlist/coprod macros
I think it can be useful to have a syntax for representing "the rest of the list" in `hlist!`, `Coprod!`, etc

 It is tempting to use `..rest` for this syntax but that would be ambiguous with otherwise perfectly valid patterns and expressions.  I suggest `/rest` as the syntax:

```rust
// Trait for HLists with at least two elements
pub trait AtLeastTwo { }

impl<A, B, Rest> AtLeastTwo for Hlist!(A, B, /Rest) { }
```

```rust
let hlist_pat![head, /tail] = list;

assert_eq!(list, hlist![head, /tail]);
```

I can basically just copy over [what I did here](https://github.com/ExpHP/funky/blob/master/src/macros.rs) if there is interest.

ExpHP:
Closes #100

In addition to adding `...` I also did the following:

* Move the macros into their own file, which is the first module declared in `core/lib.rs`.  This is common practice to help dodge issues related to the fact that macros must be declared before they are used.

* I ported over the test structs from `funky`.  These are just a bunch of types like `A`, `B`, `C`, ... that help reduce the barrier to writing unit tests by not requiring the author to be so inventive with types from the standard library.  I missed having these during the last couple of PRs...

* Added a test to verify beyond any doubt that the macros are reexported by frunk, because... I actually was very surprised to see this happening!  Apparently this is a [special feature of `pub use crate_name::*`](https://github.com/rust-lang/rust/issues/35896#issuecomment-342986313)?


core/src/macros.rs
```diff
@@ -0,0 +1,190 @@
...
+#[macro_export]
+macro_rules! Coprod {
+    () => { $crate::coproduct::CNil };
+    (...$Rest:ty) => { $Rest };
+    ($A:ty) => { Coprod![$A,] };
+    ($A:ty, $($tok:tt)*) => {
+        $crate::coproduct::Coproduct<$A, Coprod![$($tok)*]>
+    };
+}
+
```
ExpHP: Oops, I missed the `field!` macro.  That should be moved here as well.

ExpHP:
@lloydmeta can I pull the switch on this or do you prefer having the chance to review everything?

PR #110, opened 2018-04-04
Remove redundant AsRef implementations and &'a-based trait impls.

lloydmeta:
This helps us avoid overflows when implementing our recursive traits (e.g. https://github.com/lloydmeta/frunk/pull/106#issuecomment-377927198). 

Instead, introduce a `ToRef` trait and impls that allows us to turn `&'a HCons<H, Tail>` into `HCons<&'a H, <&'a Tail as ToRef>::Output>`.

Unlike `AsRef`, we are not restricted to having to return a `&T`.

See https://github.com/lloydmeta/frunk/pull/106#issuecomment-377957019

Note I couldn't really make this work with the `Hlist!` type macro I needed to use the associated `Output` type from the tail implementation.

### DIFF
--- a/core/src/hlist.rs
+++ b/core/src/hlist.rs
@@ -799,23 +799,25 @@ where
 ///
 /// This is essentially From, but the more specific nature of it means it's more ergonomic
 /// in actual usage
-pub trait ToRef {
+pub trait ToRef<'a> {
     type Output;
-    fn to_ref(self) -> Self::Output;
+    fn to_ref(&'a self) -> Self::Output;
 }
 
-impl <'a> ToRef for &'a HNil {
+impl <'a>  ToRef<'a> for HNil {
     type Output = &'a HNil;
-    fn to_ref(self) -> Self::Output {
+    fn to_ref(&'a self) -> Self::Output {
         &HNil
     }
 
 }
 
-impl <'a, H, Tail> ToRef for &'a HCons<H, Tail>
-where &'a Tail: ToRef {
-    type Output = HCons<&'a H, <&'a Tail as ToRef>::Output>;
-    fn to_ref(self) -> Self::Output {
+impl <'a, H, Tail> ToRef<'a> for HCons<H, Tail>
+where
+    H: 'a,
+    Tail: ToRef<'a> {
+    type Output = HCons<&'a H, <Tail as ToRef<'a>>::Output>;
+    fn to_ref(&'a self) -> Self::Output {
         HCons {
             head: &self.head,
             tail: (&self.tail).to_ref()
--- a/core/src/hlist.rs
+++ b/core/src/hlist.rs
@@ -286,6 +286,25 @@ macro_rules! gen_inherent_methods {
             {
                 IntoReverse::into_reverse(self)
             }
+
+            /// Return an HList where the contents are references to
+            /// the original HList on which this method was called.
+            ///
+            /// # Examples
+            ///
+            /// ```
+            /// # #[macro_use] extern crate frunk_core; fn main() {
+            /// assert_eq!(hlist![].to_ref(), hlist![]);
+            ///
+            /// assert_eq!(hlist![1, true].to_ref(), hlist![&1, &true]);
+            /// # }
+            /// ```
+            #[inline(always)]
+            pub fn to_ref<'a>(&'a self) -> <Self as ToRef<'a>>::Output
+                where Self: ToRef<'a>,
+            {
+                ToRef::to_ref(self)
+            }
         }
     };
 }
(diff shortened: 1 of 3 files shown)


core/src/hlist.rs
```diff
@@ -839,6 +794,36 @@ where
...
+///
+/// This is essentially From, but the more specific nature of it means it's more ergonomic
+/// in actual usage
+pub trait ToRef {
+    type Output;
+    fn to_ref(self) -> Self::Output;
+}
+
+impl <'a> ToRef for &'a HNil {
+    type Output = &'a HNil;
```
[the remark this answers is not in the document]
ExpHP: Ah, the examples worked because e.g. `impl HMappable<F> for &'a HNil ` are still there. I think those should go.

core/src/hlist.rs
```diff
@@ -839,6 +794,36 @@ where
     }
 }
 
+/// An alternative to AsRef that lets us create implementations for our recursive
+/// traits that takes references, without having to deal with weird impl overflows
+///
+/// This is essentially From, but the more specific nature of it means it's more ergonomic
+/// in actual usage
+pub trait ToRef {
```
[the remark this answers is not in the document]
lloydmeta: Excellent idea. Done.

core/src/hlist.rs
```diff
@@ -839,6 +794,36 @@ where
     }
 }
 
+/// An alternative to AsRef that lets us create implementations for our recursive
+/// traits that takes references, without having to deal with weird impl overflows
```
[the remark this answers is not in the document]
lloydmeta: Hah! Sorry about that; was in a rush. I've made them more formal, and linked to an example to show where the motivation comes from.

core/src/hlist.rs
```diff
@@ -839,6 +794,36 @@ where
...
+///
+/// This is essentially From, but the more specific nature of it means it's more ergonomic
+/// in actual usage
+pub trait ToRef {
+    type Output;
+    fn to_ref(self) -> Self::Output;
```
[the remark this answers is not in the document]
lloydmeta: Good idea. Added.

core/src/hlist.rs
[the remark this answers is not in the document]
lloydmeta: True; good idea :) Done.

core/src/hlist.rs
```diff
@@ -1272,17 +1247,17 @@ mod tests {
...
-    //    #[test]
-    //    fn test_foldr_non_consuming() {
-    //        let h = hlist![1, false, 42f32];
-    //        let folder = hlist![|&i, acc| i + acc,
-    //                                             |&_, acc| if acc > 42f32 { 9000 } else { 0 },
-    //                                             |&f, acc| f + acc];
-    //        let folded = h.to_ref().foldr(folder,
-    //                                      1f32);
-    //        assert_eq!(folded, 9001)
-    //    }
```
ExpHP: Glad to see this test enabled =D

core/src/hlist.rs
```diff
@@ -794,75 +805,47 @@ where
...
 }
 
-impl <'a, H, Tail> ToRef for &'a HCons<H, Tail>
-where &'a Tail: ToRef {
-    type Output = HCons<&'a H, <&'a Tail as ToRef>::Output>;
-    fn to_ref(self) -> Self::Output {
+impl<'a, H, Tail> ToRef<'a> for HCons<H, Tail>
+    where
+        H: 'a,
+        Tail: ToRef<'a>,
```
ExpHP: I can tell you use clion/intellij from how these where bounds are getting indented =P

core/src/hlist.rs
```diff
@@ -794,75 +805,47 @@ where
...
+///
+/// This lets us create implementations for our recursive traits that take the resulting
+/// Output reference type, without having to deal with strange, spurious overflows
+/// that sometimes occur when trying to implement a trait for &'a T (see this comment:
+/// https://github.com/lloydmeta/frunk/pull/106#issuecomment-377927198)
 ///
 /// This is essentially From, but the more specific nature of it means it's more ergonomic
-/// in actual usage
-pub trait ToRef {
+/// in actual usage.
```
ExpHP: I might like these docs to link to the inherent method as well.  I would follow the example of `IntoReverse` (which does not describe itself as just an implementation detail), since I think it would be nice to implement this for Coproduct as well at some point later.

lloydmeta: That's a great idea :)

core/src/hlist.rs
lloydmeta: Fixed! Oddly enough, I `cargo +nightly fmt`ed and copy and pasted the formatted new stuff, but IntelliJ was insistent on screwing it up.

ExpHP: Yep, can't trust copy and paste in intellij.  I often find myself fixing the same where bounds multiple times thanks to this. =P

PR #113, opened 2018-04-06
Add Functor and Applicative

bjorn3:
This works on stable, but is ugly

### DIFF
--- a/src/applicative.rs
+++ b/src/applicative.rs
@@ -1,0 +1,37 @@
+use kind::Kind;
+use functor::Functor;
+
+pub trait Applicative<B, F>: Functor<B> + Kind<F> + Sized
+    where F: Fn(Self::Item) -> B {
+    fn pure_(Self::Item) -> Self;
+    fn apply(f: <Self as Kind<F>>::Make, Self) -> <Self as Kind<B>>::Make;
+}
+
+impl<A, B, F> Applicative<B, F> for Option<A>
+    where F: Fn(A) -> B {
+    fn pure_(a: A) -> Option<A> {
+        Some(a)
+    }
+
+    fn apply(f: Option<F>, to: Option<A>) -> Option<B> {
+        match f {
+            Some(f) => to.map(f),
+            None => None,
+        }
+    }
+}
+
+impl<A: Clone, B, F> Applicative<B, F> for Vec<A>
+    where F: Fn(A) -> B {
+    fn pure_(a: A) -> Vec<A> {
+        vec![a]
+    }
+
+    fn apply(f: Vec<F>, to: Vec<A>) -> Vec<B> {
+        f.into_iter().flat_map(|f| {
+            to.clone().into_iter().map(|to| {
+                f(to)
+            }).collect::<Vec<B>>()
+        }).collect()
+    }
+}
--- a/src/functor.rs
+++ b/src/functor.rs
@@ -1,0 +1,27 @@
+use kind::Kind;
+
+pub trait Functor<B>: Kind<B> {
+    type Item;
+    fn fmap<F>(self, f: F) -> <Self as Kind<B>>::Make where F: Fn(Self::Item) -> B;
+}
+
+impl<A, B> Functor<B> for Option<A> {
+    type Item = A;
+    fn fmap<F>(self, f: F) -> Option<B> where F: Fn(A) -> B {
+        self.map(f)
+    }
+}
+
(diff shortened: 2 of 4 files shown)


src/applicative.rs
```diff
@@ -0,0 +1,37 @@
+use kind::Kind;
+use functor::Functor;
+
+pub trait Applicative<B, F>: Functor<B> + Kind<F> + Sized
+    where F: Fn(Self::Item) -> B {
+    fn pure_(Self::Item) -> Self;
```
ExpHP: yuck on the hack for avoiding the keyword.  I might suggest looking to other langauges for naming precedent.  (I know a lot of JS libs call this `of`; I don't know about other languages)

ISTM you are also going to immediately run into problems with type inference when you try to use this method, because its signature does not constrain `F`.

ExpHP:
I think this really needs examples of usage to demonstrate that it is useful.  Or rather, *usable.*  Unsurprisingly, 99% of my own attempts to encode HKT-like abstractions into rust only seemed like a good idea up until the point that I actually tried to *use* them.

From past experience with type level programming in rust, I doubt it is possible to get very far at all with having the element type be a type argument of `Functor`.  Typically I usually end up running into some thing that I can't write an impl for without the compiler calling it ambiguous (since the compiler doesn't know that `T: Functor<A>` implies `T: !Functor<B>`), and the only solution ends up being to turn the element type into an associated type... which obviously defeats the entire purpose of the abstraction.

In summary: I don't think the language is ready for this.

Centril:
My view is that we'd like to experiment with this once GATs land, so I agree with @ExpHP.

Centril:
Another heads up: given that Rust has affine typing and that we have `Fn`, `FnMut`, `FnOnce` there would either need to be (endo)functors for different categories, or we would need some form of `ConstraintKinds` to generalize over `FnWhatever`. I recently started an RFC-writing-project to that effect: https://github.com/Centril/rfc-trait-parametric-polymorphism/

See also the following blog post on linear monads in Haskell: https://m0ar.github.io/safe-streaming/2017/07/20/homegrown-linear-monads.html

lloydmeta:
FWIW, I think this is a really cool attempt at implementing HKTs 😎 

The `Kind<A>` with a `Make` associated type is quite clever and clean!

Just for comparisons' sake:

1. In [Arrow-kt](http://arrow-kt.io) (FP lib for Kotlin), the authors introduced a form of HKT using something like `Kind<F,A>`, where for a given type constructor, say `Option<A>`, they create a corresponding private dummy class called `ForOption`, then make the type `Option<A>` inherit from `Kind<ForOption, A>` (more details [here](http://arrow-kt.io/docs/patterns/glossary/#higher-kinds))
2. In [Funland/funfix](https://github.com/funfix/funfix) (FP lib for JS), they do something similar w/ `HK<F, A>` (see [here](https://github.com/funfix/funland/blob/e6b515c1c842277602face1e323b03c95049b06e/packages/funland/src/kinds.ts#L67-L70))

I believe both of these are based on the [Lightweight HKTs paper](https://www.cl.cam.ac.uk/~jdy22/papers/lightweight-higher-kinded-polymorphism.pdf) that showed how to emulate HKTs in OCaml.

Having said that, I think @ExpHP is right; this would be a lot nicer if you showed usage (e.g. tests or doctests) so it would be easier to see how this works out in practice :)

ExpHP:
> they create a corresponding private dummy class called `ForOption`

```rust
type ForOption = Option<()>;
```

:duck:

src/applicative.rs
Centril: Some "insider info:" `pure` will likely be removed as a keyword soon-ish.

For now, I'm partial to `pointed` or `wrap`.

lloydmeta:
Yeah, regrettably, I think it might be best to wait for GATs (which actually shouldn't be all that far off ?? (https://github.com/rust-lang-nursery/chalk/issues/116)

rami3l:
GAT is finally here after a long wait! 🍾 Time to reconsider this one?

ExpHP:
Ah, I see, so they've been stabilized on nightly.  That's good news!

I wouldn't exactly say that they're "here" yet (IIRC the `!` type has been "stabilized" only to get unstabilized again during beta, *multiple* times).  But now is definitely a good time to start playing around with the feature and helping to put it to the test, and frunk is certainly the kind of crate that could probably stand to benefit from using GATs.

So yeah, if people want to experiment with this and make PRs, I'd say go right ahead.

PR #115, opened 2018-04-08
The Great Glob Import Extinction (and requisite finishing touches)

ExpHP:
As before, I tried to break this up into individually readible commits.

It includes several of the last-minute changes [mentioned here](https://github.com/lloydmeta/frunk/issues/101#issuecomment-379569647), each as its own commit (or two) near the beginning.  This is followed by *one really big commit* that corrects all of the examples and tests to use the new `use` paths.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Statically typed heterogeneous lists.
 First, let's enable `hlist`:
 ```rust
 #[macro_use] extern crate frunk; // allows us to use the handy hlist! macro
-use frunk_core::hlist::*;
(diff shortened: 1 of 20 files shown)


ExpHP:
aaaaaahhhhh waitasecond that's not the right branch, hang on >_<

ExpHP:
So... that WAS the right branch, but github apparently lists commits on both this page and the Commits tab in chronological order (which is the wrong order, because I moved them around while rebasing). >_<

fixed with `rebase --ignore-date` in another PR (GitHub won't let me reopen this one)

PR #116, opened 2018-04-08
The Great Glob Import Extinction (and requisite finishing touches)

### LINKED ISSUE #101: Meta-issue for ergonomic advancements

ExpHP:
As before, I tried to break this up into individually readible commits.

It includes several of the last-minute breaking changes [mentioned here](https://github.com/lloydmeta/frunk/issues/101#issuecomment-379569647), each as its own commit (or two) near the beginning. This is followed by one really big commit that corrects all of the examples and tests to use the new use paths.

Closes #101

### DIFF
--- a/core/src/hlist.rs
+++ b/core/src/hlist.rs
@@ -56,6 +56,8 @@
 //! # }
 //! ```
 
+use indices::{Here, There, Suffixed};
+
 use std::ops::Add;
 
 /// Typeclass for HList-y behaviour
@@ -638,9 +640,6 @@ where
         }
     }
 }
-
-// FIXME kill this one examples are updated
-pub use ::indices::{Here, There, Suffixed};
 
 /// Trait for borrowing an HList element by type
 ///
--- a/core/src/hlist.rs
+++ b/core/src/hlist.rs
@@ -308,6 +308,150 @@ macro_rules! gen_inherent_methods {
             {
                 ToRef::to_ref(self)
             }
+
+            /// (XXX from trait XXX)
+            ///
+            /// Trait that allow for mapping over a data structure using mapping functions stored in another
+            /// data structure
+            ///
+            /// It might be a good idea to try to re-write these using the foldr variants, but it's a
+            /// wee-bit more complicated.
+            ///
+            /// (XXX from method XXX)
+            ///
+            /// Maps over the current data structure using functions stored in another
+            /// data structure.
+            ///
+            /// # Examples
+            ///
+            /// ```
+            /// # #[macro_use] extern crate frunk_core; use frunk_core::hlist::*; fn main() {
+            /// let nil = HNil;
+            ///
+            /// assert_eq!(nil.map(HNil), HNil);
+            ///
+            /// let h = hlist![1, false, 42f32];
+            ///
+            /// // Sadly we need to help the compiler understand the bool type in our mapper
+            ///
+            /// let mapped = h.to_ref().map(hlist![
+            ///     |&n| n + 1,
+            ///     |b: &bool| !b,
+            ///     |&f| f + 1f32]);
+            /// assert_eq!(mapped, hlist![2, true, 43f32]);
+            ///
+            /// // There is also a value-consuming version that passes values to your functions
+            /// // instead of just references:
+            ///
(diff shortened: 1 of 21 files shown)


ExpHP:
wtf why is 'doc updates for Func' listed after TGGIE, I thought I fixed all of the timestamps >_<

whatever, at least that's the only one out of order this time.

lloydmeta:
My simple mind is saying just a `poly.rs` module in frunk_core might work? Open to other ideas of course :)

core/src/indices.rs
```diff
@@ -0,0 +1,33 @@
...
+use std::marker::PhantomData;
+
+// Largely lifted from https://github.com/Sgeo/hlist/blob/master/src/lib.rs#L30
+
+/// Used as an index into an `HList`.
+///
+/// `Here` is 0, pointing to the head of the HList.
+///
+/// Users should normally allow type inference to create this type.
+pub struct Here(());
```
Centril: Should probably use `{ priv: () }` here to make the privacy thingy more intentional-looking.

ExpHP:
I considered it and am open to changing it, but note that there is an
observable difference to the user between a regular struct with a private
member and a tuple struct with a private member. (the trouble is, I'm not
actually sure at this point which is better for reified indices, or if it
even matters at all!)

ExpHP:
Scratch that, actually. Even tuple structs can use struct expression syntax
and struct pattern syntax, so I guess regular structs are the clear winner
for maximum flexibility in the future. I will change both Here and There.

ExpHP:
Oops, mail replies don't appear where I thought they would on this page. =P

ExpHP:
> My simple mind is saying just a poly.rs module in frunk_core might work? Open to other ideas of course :)

This is definitely possible, although I don't think `ToRef` will quite fit in there.  My trouble is with having too many little modules for small groups of things; I feel this harms discoverability.

I tried making them live at the root but I didn't like how the resulting documentation for `frunk_core/index.html` looked (it made them look too important); so for now I've moved them into a public module called `traits`.  This name feels vague enough to be able to cover most of the things I want to remove from HList; however, it goes counter to the naming scheme of the rest of frunk (which mostly groups things by feature).

core/src/hlist.rs
```diff
@@ -331,7 +331,7 @@ macro_rules! gen_inherent_methods {
             /// * A single closure (for mapping an HList that is homogenous).
             /// * A single [`Poly`].
             ///
-            /// [`Poly`]: struct.Poly.html
+            /// [`Poly`]: ../traits/struct.Poly.html
```
lloydmeta: Marginally unrelated question: are these checked by the doc-generator?

ExpHP: I installed linkchecker from `rustc/src/tools` and use it religiously

ExpHP: To clarify, I think the process is:

```rust
git clone https://github.com/rust-lang/rust
cd rust 
git submodule update # or init, I forget
cd src/tools/linkchecker
cargo install
```

Centril: @ExpHP Oh; I feel like I just discovered ice-cream. Marvelous!

PR #117, opened 2018-04-15
Add generic::map_repr + improve documentation of the module.

Centril:
Pokes back at @ExpHP ;)

I was a bit slow =P but here's the proposed `map_repr` and some documentation improvements.

### DIFF
--- a/core/src/generic.rs
+++ b/core/src/generic.rs
@@ -1,7 +1,7 @@
-//! This module holds the machinery behind Generic.
+//! This module holds the machinery behind `Generic`.
 //!
-//! It contains the Generic trait and some helper methods for using the Generic
-//! trait without having to use universal function call syntax.
+//! It contains the `Generic` trait and some helper methods for using the
+//! `Generic` trait without having to use universal function call syntax.
 //!
 //! # Examples
 //!
@@ -32,10 +32,10 @@
 //! let d_person: DomainPerson = frunk::convert_from(a_person); // done
 //! # }
 
-/// A trait that converts from a type to a generic representation
+/// A trait that converts from a type to a generic representation.
 ///
-/// For the most part, you should be using the derivation that is available through
-/// frunk_derive to generate instances of this trait for your types.
+/// For the most part, you should be using the derivation that is available
+/// through `frunk_derive` to generate instances of this trait for your types.
 ///
 /// # Laws
 ///
@@ -80,14 +80,14 @@ pub trait Generic {
     /// The generic representation type.
     type Repr;
 
-    /// Convert an object to its representation type `Repr`.
+    /// Convert a value to its representation type `Repr`.
     fn into(self) -> Self::Repr;
 
-    /// Convert an object's representation type `Repr` to the object's type.
+    /// Convert a value's representation type `Repr` to the values's type.
     fn from(repr: Self::Repr) -> Self;
 
-    /// Convert an object to a another type provided that they have
-    /// a compatible representation type.
+    /// Convert a value to an another type provided that they have
+    /// the same representation type.
     fn convert_from<Src>(src: Src) -> Self
     where
         Src: Generic<Repr = Self::Repr>,
@@ -97,10 +97,9 @@ pub trait Generic {
         <Self as Generic>::from(repr)
     }
 
-    /// Maps the given object of type `Self` by first transforming it to
-    /// the representation type, then applying a `mapper` function
-    /// on the representation type and finally transforming it back to
-    /// an object of type `Self`.
+    /// Maps the given value of type `Self` by first transforming it to
+    /// the representation type `Repr`, then applying a `mapper` function
+    /// on `Repr` and finally transforming it back to a value of type `Self`.
     fn map_repr<Repr, Mapper>(self, mapper: Mapper) -> Self
     where
         Self: Generic<Repr = Repr> + Sized,
@@ -118,7 +117,7 @@ where
     <Dst as Generic>::from(repr)
 }
 
-/// Given a object of type `Src`, returns its generic representation `Repr`.
+/// Given a value of type `Src`, returns its generic representation `Repr`.
 pub fn into_generic<Src, Repr>(src: Src) -> Repr
 where
     Src: Generic<Repr = Repr>,
@@ -127,7 +126,7 @@ where
 }
 
 /// Converts one type `Src` into another type `Dst` assuming they have the same
-/// generic representation.
+/// representation type `Repr`.
 pub fn convert_from<Src, Dst, Repr>(src: Src) -> Dst
 where
     Src: Generic<Repr = Repr>,
@@ -136,12 +135,12 @@ where
     <Dst as Generic>::convert_from(src)
 }
 
-/// Maps an object of the element type `Elt` using a function on the
-/// representation type `Repr` of `Elt`.
-pub fn map_repr<Elt, Repr, Mapper>(src: Elt, mapper: Mapper) -> Elt
+/// Maps a value of a given type `A` using a function on the
+/// representation type `Repr` of `A`.
+pub fn map_repr<A, Repr, Mapper>(val: A, mapper: Mapper) -> A
 where
-    Elt: Generic<Repr = Repr>,
+    A: Generic<Repr = Repr>,
     Mapper: FnOnce(Repr) -> Repr
 {
-    <Elt as Generic>::map_repr(src, mapper)
+    <A as Generic>::map_repr(val, mapper)
 }
--- a/core/src/generic.rs
+++ b/core/src/generic.rs
@@ -90,8 +90,8 @@ pub trait Generic {
     /// the same representation type.
     fn convert_from<Src>(src: Src) -> Self
     where
-        Src: Generic<Repr = Self::Repr>,
         Self: Sized,
+        Src: Generic<Repr = Self::Repr>,
     {
         let repr = <Src as Generic>::into(src);
         <Self as Generic>::from(repr)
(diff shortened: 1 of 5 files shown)


core/src/generic.rs
```diff
@@ -67,47 +77,71 @@
...
+
+/// Maps an object of the element type `Elt` using a function on the
+/// representation type `Repr` of `Elt`.
+pub fn map_repr<Elt, Repr, Mapper>(src: Elt, mapper: Mapper) -> Elt
+where
+    Elt: Generic<Repr = Repr>,
+    Mapper: FnOnce(Repr) -> Repr
+{
+    <Elt as Generic>::map_repr(src, mapper)
+}
```
ExpHP: help contribute to world peace; configure your editor to always leave a trailing newline

core/src/generic.rs
```diff
@@ -67,47 +77,71 @@
...
 
-    /// Go from Repr to something
-    fn from(r: Self::Repr) -> Self;
+    /// Convert an object's representation type `Repr` to the object's type.
+    fn from(repr: Self::Repr) -> Self;
 
-    /// From one type to another using a type with a compatible generic representation
-    fn convert_from<A>(a: A) -> Self
+    /// Convert an object to a another type provided that they have
+    /// a compatible representation type.
```
ExpHP: Maybe this should say "the same representation type."

core/src/generic.rs
```diff
@@ -67,47 +77,71 @@
...
-    A: Generic<Repr = Repr>,
+    Dst: Generic<Repr = Repr>,
 {
-    <A as Generic>::from(gen)
+    <Dst as Generic>::from(repr)
 }
 
-/// Given an A, returns its generic Representation
-pub fn into_generic<A, Repr>(a: A) -> Repr
+/// Given a object of type `Src`, returns its generic representation `Repr`.
```
ExpHP: **an** object.  (or perhaps better, "a value")

core/src/generic.rs
```diff
@@ -67,47 +77,71 @@
...
+pub fn into_generic<Src, Repr>(src: Src) -> Repr
```
ExpHP: I'm just noticing this now, but maybe Repr shouldn't be in the type parameter list here (since it's not necessary, and the type parameter list is public API).

core/src/generic.rs
```diff
@@ -67,47 +77,71 @@
...
 {
-    <A as Generic>::into(a)
+    <Src as Generic>::into(src)
 }
 
-/// Converts one type into another assuming they have the same generic Representation
-pub fn convert_from<A, B, Repr>(a: A) -> B
+/// Converts one type `Src` into another type `Dst` assuming they have the same
+/// generic representation.
+pub fn convert_from<Src, Dst, Repr>(src: Src) -> Dst
```
ExpHP: (if we do want to eliminate unnecessary type params, there's this one too, and probably more in `labelled`)

core/src/generic.rs
```diff
@@ -67,47 +77,71 @@
...
-    A: Generic<Repr = Repr>,
-    B: Generic<Repr = Repr>,
+    Src: Generic<Repr = Repr>,
+    Dst: Generic<Repr = Repr>,
 {
-    <B as Generic>::convert_from(a)
+    <Dst as Generic>::convert_from(src)
 }
+
+/// Maps an object of the element type `Elt` using a function on the
```
ExpHP: I find "element" confusing, it makes me think it is referring to an individual item in the HList.

core/src/generic.rs
Centril: Oh no; not again... I thought I fixed that once and for all :/

core/src/generic.rs
```diff
@@ -80,14 +80,14 @@ pub trait Generic {
     /// The generic representation type.
     type Repr;
 
-    /// Convert an object to its representation type `Repr`.
+    /// Convert a value to its representation type `Repr`.
     fn into(self) -> Self::Repr;
 
-    /// Convert an object's representation type `Repr` to the object's type.
+    /// Convert a value's representation type `Repr` to the values's type.
```
ExpHP: values's =P

core/src/generic.rs
```diff
@@ -80,14 +80,14 @@ pub trait Generic {
...
     fn from(repr: Self::Repr) -> Self;
 
-    /// Convert an object to a another type provided that they have
-    /// a compatible representation type.
+    /// Convert a value to an another type provided that they have
```
ExpHP: an another

core/src/labelled.rs
```diff
@@ -1,15 +1,18 @@
...
+//! about field names.
 //!
-//! Having a separate trait for LabelledGenerics gives us the freedom to derive both
-//! labelled and non-labelled generic type class instances for our types.
+//! Having a separate trait for `LabelledGeneric`s gives us the freedom to
+//! derive both labelled and non-labelled generic trait instances for our types.
 //!
-//! Asides from the main LabelledGeneric trait, this module holds helper methods that allow
-//! users to use LabelledGeneric without using universal function call syntax.
+//! Asides from the main `LabelledGeneric` trait, this module holds helper
```
ExpHP: just spotted this old typo "Asides from"

core/src/labelled.rs
```diff
@@ -143,48 +148,51 @@ pub trait LabelledGeneric {
...
         // We toss away the remainder.
-        let (self_gen, _): (<Self as LabelledGeneric>::Repr, _) = a_gen.sculpt();
+        let (self_gen, _): (<Self as LabelledGeneric>::Repr, _) = src_gen.sculpt();
         <Self as LabelledGeneric>::from(self_gen)
     }
 }
 
-/// Given a labelled generic Representation of an A, returns A
-pub fn from_labelled_generic<A, Repr>(gen: Repr) -> A
+/// Given a labelled generic representation of an `Dst`, returns `Dst`
```
ExpHP: an Dst

Centril:
@ExpHP Yeah no, the macros would take time I didn't have ;)

Fixed most of your nits..

Also, I noticed that there are deprecated functions in `generic.rs` and `labelled.rs`. Should we remove them?

I added `map_inter` which is more flexible in some cases (when the intermediate type is known).

Re. the extra type parameters, did you want to remove them for all the free functions?

ExpHP:
Something makes me uncomfortable about the Generic impl for HLists because it is inherently different from all of the others (it's the only time that Generic is ever recursive).  If anything I would've expected Generic for HLists to be the identity transform.  I think we should maybe hold back on that and discuss it

Centril:
@ExpHP I initially wrote `type Repr = Self;`. That also works for me. =) I'll wait for @lloydmeta's review first tho before changing.

ExpHP:
P.S. I have no idea what `intern` is supposed to be short for, but I'm also half asleep right now

Centril:
@ExpHP Intermediate ;) I was really trying to find a better name but it was really hard 😆 I'd love a better name.

core/src/generic.rs
```diff
@@ -32,10 +32,20 @@
...
-/// For the most part, you should be using the derivation that is available through
-/// frunk_derive to generate instances of this typeclass for your types.
+/// For the most part, you should be using the derivation that is available
+/// through `frunk_derive` to generate instances of this trait for your types.
+///
+/// # Laws
+///
+/// Any implementation of `Generic` must satisfy the following two laws:
+///
+/// 1. `forall x : Self. x == Generic::from(Generic::into(x))`
```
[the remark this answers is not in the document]
Centril: Yeah probably :)

Centril:
@lloydmeta So what are your thought on the `Generic` impl for `HCons`?

lloydmeta:
@Centril, the `Generic` impl looks fine to me. If anything at all, I was wondering about the `H: Generic` bound and whether it would make sense to provide `Generic` impls for primitive types 🤔

ExpHP:
Providing identity Generic impls for standard types seems like a big rabbit hole.

I just took a look at [what Haskell does](http://hackage.haskell.org/package/base-4.11.0.0/docs/GHC-Generics.html#t::-42-:) and I'm... not actually sure *what* they do.  None of the Generic impls have any constraints, including the ones most closely corresponding to this:

```haskell
instance Generic1 (f :*: g :: k -> *)
instance Generic ((f :*: g) p)
```

which seems to suggest that the transformation for `:*:` is either the identity transform or simply one that adds some extra wrapper types for metadata. But *all* of the instances are generated with DeriveGeneric (*including* the ones for `:*:`), and the `Rep` types aren't documented, so I don't have much to go on. :confused:

---

My preference still is to take no action until a very strong motivation for adding `impl Generic for HCons` arises (not just "because we can," but rather "because we *need to*"); this will give us a better understanding of why a certain definition is the right one.

Centril:
Your take no action argument makes sense to me; I'll remove the impls for now (tomorrow I mean).

So this is what `Rep ((f :*: g) p)` is defined as:
```haskell
type Rep ((f :*: g) p) =
  D1
    (MetaData ":*:" "GHC.Generics" "base" False)
    (C1
      (MetaCons ":*:" (InfixI RightAssociative 6) False)
      (S1
        (MetaSel
          (Nothing :: Maybe Symbol)
          NoSourceUnpackedness
          NoSourceStrictness
          DecidedLazy
        )
        (Rec0 (f p))
        :*:
        S1
          (MetaSel
            (Nothing :: Maybe Symbol)
            NoSourceUnpackedness
            NoSourceStrictness
            DecidedLazy
          )
          (Rec0 (g p))
       )
     )
```

lloydmeta:
> My preference still is to take no action until a very strong motivation for adding impl Generic for HCons arises (not just "because we can," but rather "because we need to"); this will give us a better understanding of why a certain definition is the right one.

Yeah, letting the need drive the implementation sounds like a solid idea :+1:

Centril:
Removed the Generic for HList impls now =) @ExpHP merge if you're comfortable with the changes now?

ExpHP:
:shipit:

PR #130, opened 2018-10-04
add Selector::get_mut

### LINKED ISSUE #129: Add Selector::get_mut
Thanks for the cool lib, I'm using it heavily to create a typesafe dataframe-like lib. 

I have a need for a `get_mut` function on the `Selector` trait. Is there a reason this is missing/would you accept a PR?

jesskfullwood:
closes #129

### DIFF
--- a/core/src/hlist.rs
+++ b/core/src/hlist.rs
@@ -560,10 +560,10 @@ impl<Head, Tail> HCons<Head, Tail> {
     /// # #[macro_use] extern crate frunk; fn main() {
     /// let mut h = hlist![1i32, true];
     ///
-    /// // type inference ensures we fetch the correct type
+    /// // Type inference ensures we fetch the correct type.
     /// *h.get_mut() = false;
     /// *h.get_mut() = 2;
-    /// // *h.get_mut() = "neigh";  // won't compile
+    /// // *h.get_mut() = "neigh";  // Won't compile.
     ///
     /// assert_eq!(h, hlist![2i32, false]);
     /// # }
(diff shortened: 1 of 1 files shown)


core/src/hlist.rs
```diff
@@ -552,6 +552,30 @@ impl<Head, Tail> HCons<Head, Tail> {
...
 
+    /// Mutably borrow an element by type from an HList.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # #[macro_use] extern crate frunk; fn main() {
+    /// let mut h = hlist![1i32, true];
+    ///
+    /// // type inference ensures we fetch the correct type
```
Centril: Start sentence with capital T and end with punctuation.

core/src/hlist.rs
```diff
@@ -552,6 +552,30 @@ impl<Head, Tail> HCons<Head, Tail> {
...
+    /// *h.get_mut() = false;
+    /// *h.get_mut() = 2;
+    /// // *h.get_mut() = "neigh";  // won't compile
```
Centril: Same here.

Centril:
It would also be nice to add some tests.

core/src/hlist.rs
Centril: I had similar nits in #124 =P

I think we should follow standard AmE writing per [RFC 1574](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md) including proper punctuation :)

ExpHP: Well, it says "all documentation." I don't regard comments as documentation, and I honestly think `// Won't compile` at the end of a line looks hilarious, even moreso if you follow it up with a period.

Anyways, it doesn't bother me that much. Just warning you that you'll have to remind me, because I *will* forget. :stuck_out_tongue:

PR #134, opened 2018-11-20
Adds ToMut trait.

remexre:
Basically the same as ToRef, except with `&mut` instead of `&`.

### DIFF
--- a/core/src/coproduct.rs
+++ b/core/src/coproduct.rs
@@ -73,7 +73,7 @@
 
 use hlist::{HCons, HNil};
 use indices::{Here, There};
-use traits::{Func, Poly, ToRef};
+use traits::{Func, Poly, ToMut, ToRef};
 
 /// Enum type representing a Coproduct. Think of this as a Result, but capable
 /// of supporting any arbitrary number of types instead of just 2.
@@ -476,6 +476,30 @@ impl<Head, Tail> Coproduct<Head, Tail> {
         ToRef::to_ref(self)
     }
 
+    /// Borrow each variant of the Coproduct mutably.
+    ///
+    /// # Example
+    ///
+    /// Composing with `subset` to match a subset of variants without
+    /// consuming the coproduct:
+    ///
+    /// ```
+    /// # #[macro_use] extern crate frunk; fn main() {
+    /// use frunk::Coproduct;
+    ///
+    /// let co: Coprod!(i32, bool, String) = Coproduct::inject(true);
+    ///
+    /// assert!(co.to_mut().subset::<Coprod!(&mut bool, &mut String), _>().is_ok());
+    /// # }
+    /// ```
+    #[inline(always)]
+    pub fn to_mut<'a>(&'a mut self) -> <Self as ToMut<'a>>::Output
+    where
+        Self: ToMut<'a>,
+    {
+        ToMut::to_mut(self)
+    }
+
     /// Use functions to transform a Coproduct into a single value.
     ///
     /// A variety of types are supported for the `Folder` argument:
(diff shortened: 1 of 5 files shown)


core/src/coproduct.rs
```diff
@@ -476,6 +476,30 @@ impl<Head, Tail> Coproduct<Head, Tail> {
         ToRef::to_ref(self)
     }
 
+    /// Borrow each variant of the Coproduct mutably.
```
Centril:
```suggestion
    /// Borrow each variant of the `Coproduct` mutably.
```

core/src/hlist.rs
```diff
@@ -324,6 +324,25 @@ macro_rules! gen_inherent_methods {
                 ToRef::to_ref(self)
             }
 
+            /// Return an HList where the contents are mutable references
```
Centril:
```suggestion
            /// Return an `HList` where the contents are mutable references
```

core/src/hlist.rs
```diff
@@ -324,6 +324,25 @@ macro_rules! gen_inherent_methods {
...
+            /// to the original HList on which this method was called.
```
Centril:
```suggestion
            /// to the original `HList` on which this method was called.
```

core/src/hlist.rs
```diff
@@ -324,6 +324,25 @@ macro_rules! gen_inherent_methods {
...
+            /// ```
+            /// # #[macro_use] extern crate frunk; fn main() {
+            /// assert_eq!(hlist![].to_mut(), hlist![]);
+            ///
+            /// assert_eq!(hlist![1, true].to_mut(), hlist![&mut 1, &mut true]);
+            /// # }
+            /// ```
+            #[inline(always)]
+            pub fn to_mut<'a>(&'a mut self) -> <Self as ToMut<'a>>::Output
+                where Self: ToMut<'a>,
```
Centril: Inconsistent style as compared to other places in this PR; use the style you've used before.

core/src/traits.rs
```diff
@@ -18,6 +18,17 @@ pub trait ToRef<'a> {
     fn to_ref(&'a self) -> Self::Output;
 }
 
+/// An alternative to AsMut that does not force the reference type to be a pointer itself.
```
Centril:
```suggestion
/// An alternative to `AsMut` that does not force the reference type to be a pointer itself.
```

core/src/traits.rs
```diff
@@ -18,6 +18,17 @@ pub trait ToRef<'a> {
...
+///
+/// This parallels [ToRef]; see it for more information.
```
Centril:
```suggestion
/// This parallels [`ToRef`]; see it for more information.
```

core/src/traits.rs
```diff
@@ -18,6 +18,17 @@ pub trait ToRef<'a> {
...
+///
+/// [ToRef]: trait.ToRef.html
```
Centril:
```suggestion
/// [`ToRef`]: trait.ToRef.html
```

remexre:
The style's largely a result of copy-pasting the equivalent ToRef parts; should I change those too?

Centril:
@remexre I think we can do that as a follow up rustfmt-only PR for cleaner diffs or as part of the release procedure.

remexre:
Looks like it's stalling out on nightly on Travis; works on my machine (on nightly) though.

Centril:
@lloydmeta Feels generally easiest to keep track of things if each PR adds to Unreleased.

lloydmeta:
@remexre would you mind adding a quick summary of this change to [the Changelog](https://github.com/lloydmeta/frunk/blob/master/CHANGELOG.md)?

remexre:
Should be done.

PR #136, opened 2018-11-21
[DEBUG] Try fixing nightly compile

lloydmeta:
Try disabling `codegen-units = 1` to speed up benches.

This is probably timing out now because of the massively-nested bench that came in via #124


lloydmeta:
Well spotted! Interesting...this used to work at one point; I wonder what happened o_o

PR #137, opened 2018-11-21
Adds FuncMut and PolyMut, which parallel Func and Poly.

remexre:
This also makes `HMappable<F>` accept `FnMut` instead of just `Fn`; I don't believe this would be breaking, but I can split it out if it is.

### DIFF
--- a/core/src/traits.rs
+++ b/core/src/traits.rs
@@ -58,6 +58,13 @@ pub trait IntoReverse {
 #[derive(Debug, Copy, Clone, Default)]
 pub struct Poly<T>(pub T);
 
+/// A wrapper like [`Poly`] for [`FuncMut`].
+///
+/// [`FuncMut`]: trait.FuncMut.html
+/// [`Poly`]: trait.Poly.html
+#[derive(Debug, Copy, Clone, Default)]
+pub struct PolyMut<T>(pub T);
+
 /// This is a simple, user-implementable alternative to `Fn`.
 ///
 /// Might not be necessary if/when Fn(Once, Mut) traits are implementable
@@ -73,4 +80,14 @@ pub trait Func<Input> {
     /// small fraction of use-cases, but it also comes at great expanse to the other 95% of
     /// use cases.
     fn call(i: Input) -> Self::Output;
+}
+
+/// This is a user-implementable alternative to `FnMut`.
+///
+/// Not necessary if/when the FnMut trait is implementable in stable Rust.
+pub trait FuncMut<Input> {
+    type Output;
+
+    /// Call the `FuncMut`.
+    fn call(&mut self, i: Input) -> Self::Output;
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]:
 - Added `ToMut` trait, which allows borrowing mutably from a Coproduct or HList.
+- Added `FuncMut` and `PolyMut`, which parallel `Func` and `Poly`.
 
 ## [0.2.2] - 2018-10-21
 - Added support for [transmogrifying (recursively sculpting)](https://docs.rs/frunk/0.2.2/frunk/labelled/trait.Transmogrifier.html) one data type into another
--- a/core/src/coproduct.rs
+++ b/core/src/coproduct.rs
@@ -73,7 +73,7 @@
 
 use hlist::{HCons, HNil};
 use indices::{Here, There};
-use traits::{Func, Poly, ToMut, ToRef};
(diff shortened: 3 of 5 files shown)


remexre:
I'm using this so I can call a trait method on an HList of values that all belong to the same trait. The specific code is here:

https://github.com/remexre/csci5607-game/blob/75103f761dba59cbe2f67980cadf53a94df5e4ca/src/main.rs#L71-L73

https://github.com/remexre/csci5607-game/blob/75103f761dba59cbe2f67980cadf53a94df5e4ca/src/lib.rs#L53-L75

There's not really a huge reason to have `PolyMut<F> where F: FuncMut` over `PolyMut<F> where F: FnMut`, though, since I should be able to make `SystemStepper` into a function that returns `impl 'a + FnMut` instead of a struct.

Centril:
+ Generalizing to `FnMut` seems good; the order of calls don't seem confusing wrt. mutation.
+ We should add `poly_fn_mut!` to consistency?

ExpHP:
The question of consistency is currently a null point; Func takes no context, so `poly_fn!` provides no way to describe context (which is a fine limitation since it is difficult to do ergonomically).  

Furthermore, @remexre's use case is a generic closure, which is *even more* monumentally difficult to make a comfortable interface around. It's only one step away from the most complicated example I have in my own [experimental repo](
https://github.com/ExpHP/unbox-macro/blob/proc-macro/examples/examples.rs)

```rust
unbox!{
    // Define a generic context
    pub struct Contains<'a, T>(&'a [T]) where T: 'a;

    // Make it a generic closure
    impl Fn <'b, U: 'b>(&self, u: &'b U) -> bool
    where T: PartialEq<U>,
    { self.0.iter().any(|t| t == u) }
}
```

(remexre's example is almost as complicated, but with a monomorphic context)

---

Honestly, in my own projects, I just write my own traits now for mapping over hlists in particular ways with certain types of context, since there's no generic solution I've found comfortable to use.

ExpHP:
Here are my recent thoughts:

* If you've tried it and it works well for you, `FuncMut` and `PolyMut` are probably fine to add.  My reasons for being against it are probably letting perfect be the enemy of good, and this is not normally the bar that `frunk` holds its features to.  There are plenty of experimental features in `frunk` like transmogrification, which have nasty edge cases that are difficult or impossible to solve.

  If `FuncMut/PolyMut` isn't good enough for somebody's use case, they can create an issue and we can iterate on the design.
* For consistency, `Func` should be revised to take a `&self` context in the next major version bump.
* I stand firm on the position that `poly_fn_mut!` is unnecessary and too difficult to implement.
* I'll go through and do a code review.

core/src/traits.rs
```diff
@@ -74,3 +81,45 @@ pub trait Func<Input> {
...
 }
+
+/// This is a user-implementable alternative to `FnMut`.
+///
+/// Not necessary if/when the FnMut trait is implementable in stable Rust.
+///
+/// ```
+/// # #[macro_use]
+/// # extern crate frunk;
+/// # use frunk::{FuncMut, PolyMut};
```
ExpHP: Because some items are exported under multiple paths, I've been making it a point for doc examples to always show the intended paths for using stuff from `frunk`, so unhide this.

Likewise for the `#[macro_use] extern crate`

core/src/traits.rs
```diff
@@ -74,3 +81,45 @@ pub trait Func<Input> {
...
+/// struct ConcatDebug<'a>(&'a mut String);
+///
+/// impl<'a, T: Debug> FuncMut<T> for ConcatDebug<'a> {
+///     type Output = ();
+///     fn call(&mut self, t: T) {
+///         write!(self.0, "{:?}", t).unwrap();
+///     }
+/// }
+///
+/// fn main() {
```
ExpHP: Hide the `fn main() {`.

There's some examples of how I would format an example like this in
https://github.com/lloydmeta/frunk/blob/master/src/validated.rs

core/src/traits.rs
```diff
@@ -74,3 +81,45 @@ pub trait Func<Input> {
...
+///     let mut string = String::new();
+///
+///     {
```
ExpHP: Really can't wait for NLL...

lloydmeta:
I haven’t taken a thorough look (currently on vacation with spotty internet) but the proposed changes and feedback look reasonable to me :) 

Also :+1: for `poly_fn_mut!` perhaps being too complex to implement nicely.  It might be nice to have though, and I’m wondering now if we should have a frunk_extras module to hold things that are not-exactly-polished-but-still-useful like that sort of thing, along with `Transmogifier`, which we can iterate quickly (if needed, backward-breakingly) on as experiments.

ExpHP:
This has been sitting here for a while, should I just push the big green button?

remexre:
I need to add examples+tests still; it's finals week here so I'm gonna be a bit busy until the 19th or 20th; I should be able to do them then, though.

PR #139, opened 2018-12-05
Add generic reusable path traversals

lloydmeta:
I hesitate to use the word "lens", but this is aims to allow similar usage.

```rust
#[derive(LabelledGeneric)]
struct Dog<'a> {
    name: &'a str,
    dimensions: Dimensions,
}

#[derive(LabelledGeneric)]
struct Cat<'a> {
    name: &'a str,
    dimensions: Dimensions,
}

#[derive(LabelledGeneric)]
struct Dimensions {
    height: usize,
    width: usize,
    unit: SizeUnit,
}

#[derive(Debug, Eq, PartialEq)]
enum SizeUnit {
    Cm,
    Inch,
}

let mut dog = Dog {
    name: "Joe",
    dimensions: Dimensions {
        height: 10,
        width: 5,
        unit: SizeUnit::Inch,
    },
};

let cat = Cat {
    name: "Schmoe",
    dimensions: Dimensions {
        height: 7,
        width: 3,
        unit: SizeUnit::Cm,
    },
};

// generic, re-usable paths
let height_lens = path!(dimensions.height);
let unit_lens = path!(dimensions.unit);

assert_eq!(*height_lens.get(&dog), 10);
assert_eq!(*height_lens.get(&cat), 7);

*height_lens.get(&mut dog) = 13;
assert_eq!(*height_lens.get(&dog), 13);
```

One thing is, it doesn't look like proc macros can be used to expand to types right now, so it gets annoying to use this in functions, where we need to bubble up the type constraint for the traversing trait :(

Commits are in logical order.

# Benchmarks

Promising; 0 overhead over normal traversal from what I can see

```
running 6 tests
test lens_path_read_mut     ... bench:           0 ns/iter (+/- 0)
test lens_path_read_ref     ... bench:           0 ns/iter (+/- 0)
test lens_path_read_value   ... bench:           0 ns/iter (+/- 0)
test normal_path_read_mut   ... bench:           0 ns/iter (+/- 0)
test normal_path_read_ref   ... bench:           0 ns/iter (+/- 0)
test normal_path_read_value ... bench:           0 ns/iter (+/- 0)
```


ExpHP:
Sorry for the slow response; it's finals week and there's tons of grading for me to do!

What is the use for this? (i.e. what new possibilities does it open?). If I understand correctly, I suppose it is to allow a function to be generic over a path? (serving a purpose like class member pointers in C++?). If that is the case there should be an example demonstrating such usage (I.e. a generic function).  However that use case feels inferior to `Fn(&T) -> &T`.

Or actually, looking more closely at the examples, it looks like this helps code be general over `&` and `&mut`?  That's pretty cool but I think the docs should explain this in prose.

lloydmeta:
> Sorry for the slow response; it's finals week and there's tons of grading for me to do!

No worries at all; it's always super busy during years end; same here 😅 

> What is the use case for this? (i.e. what new possibilities does it open?). If I understand correctly, I suppose it is to allow a function to be generic over a path? (serving a purpose like class member pointers in C++?). If that is the case there should be an example demonstrating such usage (I.e. a generic function). However that use case feels inferior to `Fn(&T) -> &T`.


Yeah, basically it allows something like this:

```rust
// Given
#[derive(LabelledGeneric)]
struct Dog<'a> {
    name: &'a str,
    dimensions: Dimensions,
}

#[derive(LabelledGeneric)]
struct Cat<'a> {
    name: &'a str,
    dimensions: Dimensions,
}

let dog = Dog {
    name: "Joe",
    dimensions: Dimensions {
        height: 10,
        width: 5,
        unit: SizeUnit::Inch,
    },
};

let cat = Cat {
    name: "Schmoe",
    dimensions: Dimensions {
        height: 7,
        width: 3,
        unit: SizeUnit::Cm,
    },
};

fn is_tall<'a, T, WidthIndices, UnitIndices>(o: &'a T) -> bool
    where
        &'a T: IntoLabelledGeneric
        + PathTraverser<
            path(dimensions.height),
            WidthIndices,
            TargetValue = &'a usize,
        > + PathTraverser<
            path(dimensions.unit),
            UnitIndices,
            TargetValue = &'a SizeUnit,
        >,
{
    let height_lens = path!(dimensions.height);
    let unit_lens = path!(dimensions.unit);
    let height = *height_lens.get(o);
    match *unit_lens.get(o) {
        SizeUnit::Cm => height > 20,
        SizeUnit::Inch => height > 5,
    }
}

// no need to pass anything extra at the call site, just the object
println!("Is {} tall? {}", cat.name, is_tall(&cat));
println!("Is {} tall? {}", dog.name, is_tall(&dog));
```

Asides from freeing you from having to pass functions into said function, or declare and implement additional traits for getters, I guess it allows the function to be specific on the *shape* of the objects.

Sadly, though, because procedural **type** macros don't exist yet, we have to do this ugliness for now 🤢  :

```rust
fn is_tall<'a, T, WidthIndices, UnitIndices>(o: &'a T) -> bool
where
    &'a T: IntoLabelledGeneric
        + PathTraverser<
            Path<HCons<(d, i, m, e, n, s, i, o, n, s), Path<HCons<(h, e, i, g, h, t), HNil>>>>,
            WidthIndices,
            TargetValue = &'a usize,
        > + PathTraverser<
            Path<HCons<(d, i, m, e, n, s, i, o, n, s), Path<HCons<(u, n, i, t), HNil>>>>,
            UnitIndices,
            TargetValue = &'a SizeUnit,
        >,
{
    let height_lens = path!(dimensions.height);
    let unit_lens = path!(dimensions.unit);
    let height = *height_lens.get(o);
    match *unit_lens.get(o) {
        SizeUnit::Cm => height > 20,
        SizeUnit::Inch => height > 5,
    }
}
```

lloydmeta:
Having let this sit for a while, I'm leaning towards merging this as it

1. Doesn't have any weird edge cases
2. Fits with the theme of enabling shape-based things to happen 

Sure there aren't procedural macros for types yet, so usage in functions is a bit weird, but I think it'll come soon based on how things have progressed in the past.

PR #142, opened 2019-01-21
Allow transmogrifying through Vec

bossmc:
This is more a proof of concept than a fully fledged PR, but I wanted to get feedback early before wandering down a garden path of pain.

This PR specifically allows conversions of the form:

```
#[derive(Debug, frunk_derives::LabelledGeneric)]
struct Foo {
    bar: Vec<Bar>,
}

#[derive(Debug, frunk_derives::LabelledGeneric)]
struct Foo2 {
    bar: Vec<Bar2>,
}

#[derive(Debug, frunk_derives::LabelledGeneric)]
struct Bar {
    name: String,
}

#[derive(Debug, frunk_derives::LabelledGeneric)]
struct Bar2 {
    name: String
}

fn main() {
    let foo = Foo {
        bar: vec![
            Bar { "Sarah".to_string() },
        ]}
    };
    let foo2: Foo2 = foo.transmogrify();
    println!("{:?}", foo2);
}
```

Limitations:

* Only supports `Vec` (though other stdlib types would be easy to add)
* Only supports `Vec -> Vec`, not (e.g.) `Vec -> List` or similar (maybe judicious use of `FromIterator`?)

Does this look useful/sensible?  Have I missed anything in the implementation?

### DIFF
--- a/core/src/indices.rs
+++ b/core/src/indices.rs
@@ -53,3 +53,7 @@ pub struct LabelledGenericTransmogIndicesWrapper<T>(PhantomData<T>);
 
 /// Index type wrapper for transmogrifying a generic plucked Source to a generic Target
 pub struct PluckedLabelledGenericIndicesWrapper<T>(T);
+
+/// Index type wrapper for transmogrifying through a (known) container (e.g. `Vec`).
+pub struct MappingIndicesWrapper<T>(PhantomData<(T)>);
+
--- a/core/src/labelled.rs
+++ b/core/src/labelled.rs
@@ -690,6 +690,18 @@ pub trait Transmogrifier<Target, TransmogrifyIndexIndices> {
     fn transmogrify(self) -> Target;
 }
 
+pub struct MappingIndicesWrapper<T>(PhantomData<(T)>);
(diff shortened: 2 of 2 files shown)


ExpHP:
Re: FromIterator, for things like this I fear it becomes too easy to rule out other useful impls.  I also feel like `Vec -> HashMap` doesn't make sense.  (if there was a `HashMap -> HashMap`, I'd expect it to only operate on values, since operating on keys could create duplicates)

bossmc:
Added a test, also moved the marker index to `indices` for consistency.

I've not done anything clever (e.g. IntoIterator/FromIterator) or added support for other container types, though most would be pretty trivial.

bossmc:
Can/should this be part of https://github.com/lloydmeta/frunk/milestone/1?

lloydmeta:
@ExpHP @Centril if you'd like to review this before merging, just shout :), else I'm thinking we merge this sometime this week and push out a release this weekend :)

lloydmeta:
Released as 0.2.4, thanks again 😄

PR #148, opened 2019-02-04
#[no_std] support

### LINKED ISSUE #147: make a subset of Frunk 'no_std' compatible.
Heya! 
I work a lot on embedded no_std projects, and i think having Frunk (or rather, a subset) be no_std compatible would be very nice. 
I looked through the code, and everything using std::cmp, std::ops, marker types, std::cell, etc is available under core as well. The main culprit are the impls for monoid and semigroup, because the collections crate is usually not available.  However, there are alternatives (for example, for vec and hashmaps: https://japaric.github.io/heapless/heapless/index.html). 
I Dont think it would be that hard to split it off in a nice way as to allow impls of monoid and friends to be under a cfg flag, for example.
I'd like to help if there is some interest, would just like to know if there is a massive use case for std that i totally missed :)

ExpHP:
Closes #147

The breaking changes are:

* A change in the signature of two functions (`monoid::combine_all`, `semigroup::combine_all_option`) may break type inference
* Adding default-enabled features `std` and `validated` may break client crates using `default-features = false`.

---

The toughest part is the tests.  Several doctests were cfged out using the fact that doc comments are actually attributes.

It is crucial that `frunk_derive` does not depend on the `std` feature of `frunk_core`, or else we'd have to remove all usage of `#[derive]` from the doc tests in `frunk_core`.  This is because it is impossible to disable features of `frunk_core` enabled by `frunk_derive` due to the circular `dev-dependency`.  (features enabled by `dev-dependencies` are enabled in all builds, not just tests)

Thankfully, `frunk_derive` currently does not actually need to depend on `frunk_core` at all, so I removed the dependency.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,89 +12,90 @@
 //!   5. Semigroup
 //!   6. Monoid
 //!
-//! Here is a small taste of what Frunk has to offer:
-//!
-//! ```
-//! # #[macro_use] extern crate frunk;
-//! # #[macro_use] extern crate frunk_core;
-//! # fn main() {
-//! use frunk::prelude::*;
-//! use frunk::{self, monoid, Semigroup, Generic};
-//!
-//! // Combining Monoids
-//! let v = vec![Some(1), Some(3)];
-//! assert_eq!(monoid::combine_all(&v), Some(4));
-//!
-//! // HLists
-//! let h = hlist![1, "hi"];
-//! assert_eq!(h.len(), 2);
-//! let hlist_pat!(a, b) = h;
-//! assert_eq!(a, 1);
-//! assert_eq!(b, "hi");
-//!
-//! let h1 = hlist![Some(1), 3.3, 53i64, "hello".to_owned()];
-//! let h2 = hlist![Some(2), 1.2, 1i64, " world".to_owned()];
-//! let h3 = hlist![Some(3), 4.5, 54, "hello world".to_owned()];
-//! assert_eq!(h1.combine(&h2), h3);
-//!
-//! // Generic and LabelledGeneric-based programming
-//! // Allows Structs to play well easily with HLists
-//!
-//! #[derive(Generic, LabelledGeneric)]
-//! struct ApiUser<'a> {
-//!     FirstName: &'a str,
-//!     LastName: &'a str,
-//!     Age: usize,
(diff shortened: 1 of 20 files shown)


ExpHP:
Haven't looked at the failure yet, but the CI testing should be done differently

https://users.rust-lang.org/t/how-to-make-a-test-guarantee-that-a-crate-was-built-with-no-std/24964

ExpHP:
I expanded the initial comment.

TheVova:
Nice! I'll give this a test spin and look around to see if i can help fix some stuff as i find them :)

ExpHP:
I just remembered that I forgot about this.

IIRC the most recently merged PR was incompatible with this, so it will need a quick rebase.

ExpHP:
Oh, for crying out loud:

I am once again stumped how to make this work.  The last `frunk` update added all this proc macro stuff which uses the `proc-macro2` crate which does not appear to support `#[no_std]`.

I think more specifically the issue is that we use `proc-macro-hack`, which requires the existence of a normal crate `frunk_proc_macros` that depends on `proc-macro-hack`, which depends on `proc-macro2`.  Because `frunk_proc_macros` is not a `proc-macro` crate, cargo sees all of its dependencies as runtime dependencies of `frunk`.

ExpHP:
I'd like to merge this soon because basically every single intervening PR has broken it.

src/lib.rs
```diff
@@ -11,89 +12,90 @@
...
-//! }
-//! // let d_user = <DeletedUser as LabelledGeneric>::convert_from(s_user); <-- this would fail at compile time :)
-//!
-//! // This will, however, work, because we make use of the Sculptor type-class
-//! // to type-safely reshape the representations to align/match each other.
-//! let d_user: DeletedUser = frunk::transform_from(s_user);
-//! assert_eq!(d_user.first_name, "Joe");
-//! # }
-//! ```
+#![cfg_attr(feature = "std", doc = r#"
```
lloydmeta: Interesting !

Centril:
@lloydmeta @ExpHP Btw I should note that my language & release team duties currently prevent me from devoting any time to this crate atm. So please don't feel the need to ask for my opinions. :) (I might have time in the future...)

PR #152, opened 2019-03-15
More mapping transmogrifications

bossmc:
Add transmogrification through:

* `Box<T>`
* `Option<T>`
* `LinkedList<T>`
* `VecDeque<T>`
* `Result<T, U>`
    * Both `U` and `T` can be recursed into

The `Result` implementation is more complicated because it's possible for the `T` to be a labelled generic while the `U` isn't (or vice-versa) and hence I've added a proper `Repr` for it and a case for transmogrifying the `Repr`s based on the internals.

Unit test included for all of the above.

### DIFF
--- a/core/src/labelled.rs
+++ b/core/src/labelled.rs
@@ -425,7 +425,11 @@ where
     Type: fmt::Debug,
 {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "ValueField{{ name: {}, value: {:?} }}", self.name, self.value)
+        write!(
+            f,
+            "ValueField{{ name: {}, value: {:?} }}",
+            self.name, self.value
+        )
     }
 }
 
@@ -721,16 +725,57 @@ impl<Key, SourceValue> Transmogrifier<SourceValue, IdentityTransMog> for Field<K
     }
 }
 
-/// Implementation of `Transmogrifier` that maps over a `Vec` in a `Field`, transmogrifying the
-/// elements on the way past.
+/// Implementations of `Transmogrifier` that allow recursion through stdlib container types.
 #[cfg(feature = "std")]
+mod std {
+    use super::MappingIndicesWrapper;
+    use super::{Field, Transmogrifier};
+    use std::collections::{LinkedList, VecDeque};
+
+    macro_rules! transmogrify_seq {
+        ($container:ident) => {
+            /// Implementation of `Transmogrifier` that maps over a `$container` in a `Field`, transmogrifying the
+            /// elements on the way past.
+            impl<Key, Source, Target, InnerIndices>
+                Transmogrifier<$container<Target>, MappingIndicesWrapper<InnerIndices>>
+                for Field<Key, $container<Source>>
+            where
+                Source: Transmogrifier<Target, InnerIndices>,
+            {
+                fn transmogrify(self) -> $container<Target> {
+                    self.value.into_iter().map(|e| e.transmogrify()).collect()
+                }
+            }
+        };
+    }
+
+    transmogrify_seq!(Vec);
+    transmogrify_seq!(LinkedList);
+    transmogrify_seq!(VecDeque);
+
+    /// Implementation of `Transmogrifier` that maps over an `Box` in a `Field`, transmogrifying the
+    /// contained element on the way past.
+    impl<Key, Source, Target, InnerIndices>
+        Transmogrifier<Box<Target>, MappingIndicesWrapper<InnerIndices>> for Field<Key, Box<Source>>
+    where
+        Source: Transmogrifier<Target, InnerIndices>,
+    {
+        fn transmogrify(self) -> Box<Target> {
+            Box::new(self.value.transmogrify())
+        }
+    }
+}
+
+/// Implementation of `Transmogrifier` that maps over an `Option` in a `Field`, transmogrifying the
+/// contained element on the way past if present.
 impl<Key, Source, Target, InnerIndices>
-    Transmogrifier<Vec<Target>, MappingIndicesWrapper<InnerIndices>> for Field<Key, Vec<Source>>
+    Transmogrifier<Option<Target>, MappingIndicesWrapper<InnerIndices>>
+    for Field<Key, Option<Source>>
 where
     Source: Transmogrifier<Target, InnerIndices>,
 {
-    fn transmogrify(self) -> Vec<Target> {
-        self.value.into_iter().map(|e| e.transmogrify()).collect()
+    fn transmogrify(self) -> Option<Target> {
+        self.value.map(|e| e.transmogrify())
     }
 }
 
@@ -862,6 +907,7 @@ where
 mod tests {
     use super::chars::*;
     use super::*;
+    use std::collections::{LinkedList, VecDeque};
 
     // Set up some aliases
     #[allow(non_camel_case_types)]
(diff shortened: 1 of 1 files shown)


ExpHP:
My gut feeling (without having looked at the implementation yet) is that Result should only recurse into `T`, not the error type.  Support for one that does both sides could be provided as integration with the `either` crate.

core/src/labelled.rs
```diff
@@ -617,6 +617,31 @@ where
...
+    Err(E),
+}
+
+impl<T, E> LabelledGeneric for Result<T, E> {
+    type Repr = ResultRepr<Hlist!(Field<(), T>), Hlist!(Field<(), E>)>;
+
+    fn into(self) -> Self::Repr {
+        match self {
+            Ok(t) => ResultRepr::Ok(hlist!(field!((), t))),
+            Err(e) => ResultRepr::Err(hlist!(field!((), e))),
```
ExpHP: Why an empty string for the field name? Is there precedent for this?

core/src/labelled.rs
```diff
@@ -617,6 +617,31 @@ where
     }
 }
 
+/// A generic representation for a `Result`.  This is not simply a `Result` to provide
+/// disambiguation between conversion to `Repr` and `transmogrification()` between `Repr`s.
```
ExpHP: Can you detail what conflict you run into without this?

core/src/labelled.rs
```diff
@@ -723,15 +748,66 @@ impl<Key, SourceValue> Transmogrifier<SourceValue, IdentityTransMog> for Field<K
...
+            Source: Transmogrifier<Target, InnerIndices>,
+        {
+            fn transmogrify(self) -> $container<Target> {
+                self.value.into_iter().map(|e| e.transmogrify()).collect()
+            }
+        }
+    };
+}
+
+use std::collections::{LinkedList, VecDeque};
```
ExpHP: Yikes, more things that will require a rebase of the `#[no_std]` PR.  I'd really rather push that through first.

core/src/labelled.rs
bossmc: I don't think so, this is (as far as I know) the first `Repr` for an `enum` type.  I considered using the variant name as the field name, but that doesn't translate well to an enum like:

```
enum Foo {
  Var1(Bar, Baz),
  Var2(Quux, Xyzzy)
}
```

Maybe it should agree with the naming for tuple structs (`_0`, `_1`...)?

core/src/labelled.rs
bossmc: Fair point - what stage is that PR at?  I'm interested because at my company people really want to be able to use transmogrification in our projects but we're blocked behind needing the function added in this PR.

Is there anything I can do to help make progress on the `#[no_std]` PR to unblock this one?

core/src/labelled.rs
bossmc: There's a existing (and fairly critical) conversion for `T -> T::Repr` and I was adding a conversion from `Result::Repr -> Result::Repr`.  If `Result::Repr` was `Result` then there would be no way to disambiguate between these two conversions.

bossmc:
> My gut feeling (without having looked at the implementation yet) is that Result should only recurse into `T`, not the error type. Support for one that does both sides could be provided as integration with the `either` crate (a type specifically designed so that both type parameters play similar roles).

The use case I'm looking at is for converting between types that appear in auto-generated crates, in many cases the internal types are created from the same inputs to the auto-generation so they're "identical", but because they're in different crates they're deemed different.  These types get used in all kinds of containers within the autogenerated code (in particular, some get used as the `T` of `Result`s and some are used as the `E` of `Result`s) hence the desired to be able to transmogrify these objects wherever they appear in container types.

core/src/labelled.rs
ExpHP: It is now merged

ExpHP:
Can I see where `Result` appears in your generated code?  To me, `Result` is a control flow construct that should seldom appear inside a type (except in something ephemeral and generic like `std::iter::Peekable`).  For something that is semantically a container, one should be using `Either`.  (and if `Result` only appears as the top-level type, I'm not sure what the issue is with using `.map()` or `.map_err()` unless the calls to `.transmogrify()` are also in generated code)

core/src/labelled.rs
ExpHP: I would say go with `_0`.

bossmc:
I've had a chat with various people, and we've concluded you're correct, the `Result` instances are always the top-level type (i.e. not nested in other structures).  I can either remove the impl completely, or restrict it to only allowing the T to be recursed into.  I suspect that, given the existence of `map(|t| t.transmogrify())` you'll say remove it.

ExpHP:
I am equally fine with either option.  Having it goes well together with the `Option` impl. If you feel that others may expect it to transform the error type, however, then removing it may be the safer bet.

lloydmeta:
Given a `result: Result<T, Err1>` inside a `fn foo -> Result<T, Err2>`, when you do `let t = result?`, if there's an error, the compiler is hard-coded to do a transformation from `Err1` to `Err2` if there exists an `impl From<Err2> for Err1` right? On the surface, it feels similar to what's being done here, albeit I can also see that transmogrification might be considered more "magical" and the semantics are a bit different...

I actually don't have a good conclusion for this trail of thought, but just thought it might help advance the discussion :)

bossmc:
As a pragmatic solution I'm going to remove the `Result` support completely from this PR, as the rest seems uncontroversial, and raise a separate PR for `Result` support (and further discussion).

bossmc:
I've applied the above markups (and removed the `Result` support).

core/src/labelled.rs
```diff
@@ -425,7 +425,11 @@ where
     Type: fmt::Debug,
 {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "ValueField{{ name: {}, value: {:?} }}", self.name, self.value)
+        write!(
+            f,
+            "ValueField{{ name: {}, value: {:?} }}",
+            self.name, self.value
+        )
```
ExpHP: Is this just autoformatting?

(unrelated to this PR: now that I've noticed this, it should be using `DebugStruct`.  I'll make a separate PR for that)

bossmc: Yeah, I run rustfmt pretty much on auto-pilot 😀

bossmc:
I've not pushed the `Result` implementation as I think there's a nicer solution which I'm prototyping now.

lloydmeta:
> I've not pushed the `Result` implementation as I think there's a nicer solution which I'm prototyping now.

Nice. I wonder if we can make something generic so that the work you do there Dan be used to make it possible to derive `LabelledGeneric` for `enum`s in the general case. Perhaps we can make use of [coproduct](https://github.com/lloydmeta/frunk/blob/master/README.md#coproduct)?

Inspiration : [GHC](https://wiki.haskell.org/GHC.Generics) uses it for the equiv of enum there (Multi constructor)

bossmc:
Indeed, that's the plan, I've got the `Transmogrify` impl working and I'm making good progress on the `derive` enhancements.  I'm currently working "out of tree" by importing `frunk-proc-,macro-helpers` (so I can deliver something to my coworkers while we thrash out the upstreaming).

I didn't think to use `Coproduct` because I'm not super up on my Category Theory, so the current implementation uses 

```
enum HEither<H, T> {
  Head(H),
  Tail(T),
}
```

With a common `Repr` being `HEither<T1, HEither<T2, HEither<T3, Void>>>` (plus labels for conversion).

lloydmeta:
Nice :) That works too; that's basically all `Coproduct` really is anyway :D Sounds like you're cooking up some awesome stuff; can't wait to see it :)

bossmc:
Phew, that took longer than expected... I had a working code base weeks ago, then the real world got in the way.

To allow me to ship this function to teams at work, I've actually done the implementation in a separate pair of crates, which I've just open-sourced at https://github.com/Metaswitch/frunk-enum in case you're interested.

Issue #138, opened 2018-11-24
Add a &self context to Func


ExpHP:
If #137 gets merged, then `Func` will take no context while `FuncMut` takes a `&mut` context.  For consistency, `Func` should be revised to take a `&self` context.

IMO, `poly_fn!` still should remain limited to defining `Func`s with unit context, because it is too complicated to define an ergonomical grammar for a macro that allows the function to capture anyhing. In other words, to take advantage of this new feature will require a manual `Func` impl.

Centril:
I'm working on a proposal to add generic closures to Rust itself in any case so that seems OK. :)

ExpHP:
That sounds boss.

PR #153, opened 2019-03-21
Debug impls should use DebugStruct

ExpHP:
Just best practices; Debug impls should use the helpers on `fmt::Formatter` when possible in order to properly support `{:#?}`.

### DIFF
--- a/core/src/labelled.rs
+++ b/core/src/labelled.rs
@@ -416,7 +416,11 @@ where
     Type: fmt::Debug,
 {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "Field{{ name: {}, value: {:?} }}", self.name, self.value)
+        f.debug_struct("Field")
+            // show name without quotes
+            .field("name", &DebugAsDisplay(&self.name))
+            .field("value", &self.value)
+            .finish()
     }
 }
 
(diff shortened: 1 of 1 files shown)


lloydmeta:
I think after the conflicts in this one are fixed + merged, with #148 and #152, we have (more than) enough for a new minor-version-bumped release.

ExpHP:
This has to be a major version bump because of the change to features in the `no_std` PR.  For instance, somebody could be using Validated with `default_features = false`.

For that reason, I also want to close at least #138 which seems easy enough.

lloydmeta:
> The next release has to be a major version bump because of the change to features in the `no_std` PR. For instance, somebody could be using Validated with `default_features = false`.
> 
> For that reason, I also want to close at least #138 which seems easy enough.

🤦‍♂️ yes of course you're right; it needs to be a major bump.

lloydmeta:
Hmm, actually no, I think since we are in x.y.z where x < 0 world, we are ok with just bumping the minor, even with a breaking change, and leaving a note

* https://semver.org/
  >  Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.
* https://github.com/semver/semver/issues/333

Staying with a minor version bump seems more preferable to me than force-introducing a 1.x.x which has major implications.

ExpHP:
To be clear, that's what I meant by a major version bump.  (I meant a bump of the first nonzero part of the version number)

PR #157, opened 2019-06-20
Refactor derives

Diggsey:
This should make it much easier to extend these derives to support enums.

The only change to behaviour should be that `LabelledGeneric` can now be derived on unit structs.

Changes to the generated code:
- Renamed `'_frunk_labelled_generic_ref_` to `_frunk_ref_` because the new code is no longer specific to labelled generics. It would be easy to extend the "ref" implementations to support plain `Generic` if desired.

- Instead of extracting the `value` from `Field`s using property accessor syntax (`field.value`) it is destructed directly from the HList, using `Field { value: <x>, .. }`. This has the advantage of allowing the construction and destructuring of the user's type to use identical syntax.

Changes to naming conventions:
- There are three main "places" where generated code is used: types, expressions and patterns. This is complicated by the fact that much syntax is valid as both an expression and a pattern. The convention I used is for `build_xxx_constr` methods to be usable in both expression and pattern locations, whilst `build_xxx_type`, `build_xxx_expr` and `build_xxx_pat` methods are self explanatory.

- For normal/ref/mutable ref variations, always use the suffixes <none>/_ref/_mut.

There are several features that could be added as follow-ups to this PR:
- Addition of `IntoGeneric` trait and implementation of `Generic` for references to types deriving `Generic`.
- Support for deriving `Generic` and `LabelledGeneric` on enums.

Are these features you would want to add?

### DIFF
--- a/proc-macro-helpers/src/lib.rs
+++ b/proc-macro-helpers/src/lib.rs
@@ -19,8 +19,10 @@ use proc_macro::TokenStream;
 use proc_macro2::TokenStream as TokenStream2;
 use quote::ToTokens;
 use quote::__rt::Span;
-use syn::{DeriveInput, Expr, Ident, Member, Fields, Field, Generics, GenericParam, Lifetime, LifetimeDef};
 use syn::spanned::Spanned;
+use syn::{
+    DeriveInput, Expr, Field, Fields, GenericParam, Generics, Ident, Lifetime, LifetimeDef, Member,
+};
 
 /// These are assumed to exist as enums in frunk_core::labelled
 const ALPHA_CHARS: &'static [char] = &[
@@ -50,7 +52,7 @@ pub fn call_site_ident(s: &str) -> Ident {
 pub fn build_hlist_type<L: IntoIterator>(items: L) -> TokenStream2
 where
     L::Item: ToTokens,
-    L::IntoIter: DoubleEndedIterator
+    L::IntoIter: DoubleEndedIterator,
 {
     let mut result = quote! { ::frunk_core::hlist::HNil };
     for item in items.into_iter().rev() {
@@ -64,7 +66,7 @@ where
 pub fn build_hlist_constr<L: IntoIterator>(items: L) -> TokenStream2
 where
     L::Item: ToTokens,
-    L::IntoIter: DoubleEndedIterator
+    L::IntoIter: DoubleEndedIterator,
 {
     let mut result = quote! { ::frunk_core::hlist::HNil };
     for item in items.into_iter().rev() {
@@ -120,9 +122,10 @@ fn encode_as_ident(c: &char) -> Vec<Ident> {
 
 pub fn build_path_type(path_expr: Expr) -> impl ToTokens {
     let idents = find_idents_in_expr(path_expr);
-    idents.iter().map(|i| build_label_type(i)).fold(
-        quote!(::frunk_core::hlist::HNil),
-        |acc, t| {
+    idents
+        .iter()
+        .map(|i| build_label_type(i))
+        .fold(quote!(::frunk_core::hlist::HNil), |acc, t| {
             quote! {
             ::frunk_core::path::Path<
                 ::frunk_core::hlist::HCons<
@@ -131,8 +134,7 @@ pub fn build_path_type(path_expr: Expr) -> impl ToTokens {
                 >
               >
             }
-        },
-    )
+        })
 }
 
 /// Returns the idents in a path like expression in reverse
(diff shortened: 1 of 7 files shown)


lloydmeta:
Thanks so much for doing this. Indeed, this area of the code has been needing some additional love for a while, exactly for the reasons you cited (better readability, extendability)

> There are several features that could be added as follow-ups to this PR:
> 
> * Addition of `IntoGeneric` trait and implementation of `Generic` for references to types deriving `Generic`.
> * Support for deriving `Generic` and `LabelledGeneric` on enums.
> 
> Are these features you would want to add?

Most definitely, those would be amazing to have. If I was a prioritising man, I would say the deriving `LabelledGeneric` for enums would probably be on the top of that list :)

Also, sorry for the delayed response; I'm currently on vacation and will take a closer look when I'm back and settled in. Looking awesome thus far !

proc-macro-helpers/src/lib.rs
```diff
@@ -9,15 +9,18 @@
...
 extern crate frunk_core;
 extern crate proc_macro;
+extern crate proc_macro2;
 
 #[macro_use]
 extern crate quote;
 extern crate syn;
 
 use proc_macro::TokenStream;
+use proc_macro2::TokenStream as TokenStream2;
```
lloydmeta: Interesting; just for my own curiosity, is there any reason why we wouldn't pick one of either `proc_macro` or `proc_macro2`?

Diggsey: Honestly I'm not super familiar with all the details of how procedural macros work nowadays. The course of events was:
- I had to replace the use of `impl ToTokens` with a concrete type due to this issue with `impl Trait`: https://old.reddit.com/r/rust/comments/c2ttxg/send_and_sync_leak_through_impl_trait_because/ermiyyu/

- The concrete type returned by the `quote!` macro is from `proc_macro2`, so I had to import that.

My limited understanding is that `proc_macro::TokenStream` is tied to the compiler's plugin architecture (ie. you *can't* use `proc_macro2` in the top-level signature), and `proc_macro2::TokenStream` is used by `quote`, etc. so that the APIs can evolve separately (avoiding the issue where compiler releases would break `syn` frequently) and so both are necessary.

ExpHP: To my knowledge, it is very common for a crate to need both `proc_macro` and `proc_macro2` TokenStreams. (`proc_macro` for the entry points, `proc_macro2` for everything else).

* https://github.com/dtolnay/indoc/blob/master/impl/src/lib.rs
* https://github.com/serde-rs/serde/blob/master/serde_derive/src/de.rs

Diggsey:
I've added a rustfmt commit.

proc-macro-helpers/src/lib.rs
lloydmeta: Wow, TIL (2 things actually, that there was an **old**.reddit.com, and this leakage).  Somehow the leakage seems .... desirable to me, but I haven't really given it more than a passing thought.

> * I had to replace the use of `impl ToTokens` with a concrete type due to this issue with `impl Trait`: https://old.reddit.com/r/rust/comments/c2ttxg/send_and_sync_leak_through_impl_trait_because/ermiyyu/

Again, for my own learning: was using `impl ToTokens` as a return type causing us problems?

Diggsey: Yes, due to the implicit lifetimes, if you replace all instances of `-> TokenStream2` with `-> impl ToToken` and try to compile, you will get several errors like the following:

```
error: cannot infer an appropriate lifetime
   --> proc-macro-helpers\src\lib.rs:273:38
    |
272 |     pub fn build_hlist_type<R: ToTokens>(&self, f: impl Fn(&FieldBinding) -> R) -> impl ToTokens {
    |                                                                                    ------------- this return type evaluates to the `'static` lifetime...
273 |         build_hlist_type(self.fields.iter().map(f))
    |                          ----------- ^^^^
    |                          |
    |                          ...but this borrow...
    |
note: ...can't outlive the anonymous lifetime #1 defined on the method body at 272:5
   --> proc-macro-helpers\src\lib.rs:272:5
    |
272 | /     pub fn build_hlist_type<R: ToTokens>(&self, f: impl Fn(&FieldBinding) -> R) -> impl ToTokens {
273 | |         build_hlist_type(self.fields.iter().map(f))
274 | |     }
    | |_____^
help: you can add a constraint to the return type to make it last less than `'static` and match the anonymous lifetime #1 defined on the method body at 272:5
    |
272 |     pub fn build_hlist_type<R: ToTokens>(&self, f: impl Fn(&FieldBinding) -> R) -> impl ToTokens + '_ {
    |                                                                                    ^^^^^^^^^^^^^^^^^^
```

Diggsey: Adding the `'_` lifetime as suggested does not help: the problem here is that the lifetime of the return type is *always* `'static` regardless of the lifetime on `&self`, because `TokenStream2` has no lifetime specifiers, but there's no way to make `impl Trait` understand that.

lloydmeta: Aahhh, understood. Thanks for the answers I’ve learnt a lot :)

PR #159, opened 2019-08-12
Explore using proc_macro_hygiene

lloydmeta:
This allows us to use proc macros as types and functions without `proc_macro_hack`

This currently only compiles on `nightly` (`cargo +nightly $cmd`), and requires https://github.com/rust-lang/rust/issues/54727 (or  https://github.com/rust-lang/rust/pull/63931 ?) to be released on stable before being merged.

Screen Shot 2019-08-12 at 16 33 51

### DIFF
--- a/proc-macros/Cargo.toml
+++ b/proc-macros/Cargo.toml
@@ -11,18 +11,22 @@ keywords = ["Frunk", "macros"]
 [badges]
 travis-ci = { repository = "lloydmeta/frunk" }
 
-[dependencies.proc-macro-hack]
-version = "0.5"
+[lib]
+proc-macro = true
+
+[dependencies]
+syn = "0.15"
+quote = "0.6"
+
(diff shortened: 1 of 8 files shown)


proc-macros/Cargo.toml
```diff
@@ -11,20 +11,24 @@ keywords = ["Frunk", "macros"]
...
-[dependencies.frunk_proc_macros_impl]
-path = "../proc-macros-impl"
-default-features = false
-version = "0.0.3"
-
 [dev-dependencies.frunk]
 path = "../."
 default-features = false
-version = "0.3.0"
+version = "0.3.0"
```
ExpHP: you should configure your editor to always add a trailing newline (trailing newlines are the safest default)

lloydmeta: Good idea. I’ll look into it

lloydmeta:
Closing in favour of #161

PR #163, opened 2020-01-02
Fix wrong `frunk-core` reference in README.md

fusillicode:
Trying to use the `Generic` macro I stumbled upon the need to also import `frunk_core`.

Unfortunately using the suggestion specified inside the `README.md` lead me to the following error 😞

```
cargo check
    Updating crates.io index
error: no matching package named `frunk-core` found
location searched: registry `https://github.com/rust-lang/crates.io-index`
perhaps you meant: frunk_core
```
Using `frunk_core` however seems to work 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,19 +290,19 @@ let d_user: DeletedUser = frunk::transform_from(s_user);
 
 Sometimes you need might have one data type that is "similar in shape" to another data type, but it
 is similar _recursively_ (e.g. it has fields that are structs that have fields that are a superset of
-the fields in the target type, so they are transformable recursively).  `.transform_from` can't help you 
-there because it doesn't deal with recursion, but the `Transmogrifier` can help if both are `LabelledGeneric` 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -644,22 +644,22 @@ assert_eq!(combine_all(&product_nums), Product(24))
...
+For example, if you'd like to use just `frunk_core` with serde
 
 ```toml
 [dependencies]
-frunk-core = { version = "$version", features = ["serde"] }
+frunk_core = { version = "$version", features = ["serde"] }
 ```
 
-Or, if you'd like to use `frunk` with serde, you need to explicitly include `frunk-core` as well
+Or, if you'd like to use `frunk` with serde, you need to explicitly include `frunk_core` as well
```
lloydmeta: 🤦🏻‍♂️doh ! thanks !

fusillicode: Well, thank you for this wonderful crate! 😄 🙇‍♂️

PR #167, opened 2020-07-25
Add hlist::HAppender

Hirrolot:


lloydmeta:
Neat ! Just an FYI, there's an implementation of `Add` for HList that does essentially the same thing 🙂

https://github.com/lloydmeta/frunk/blob/941ba138f406f88a53f67f6eab3f8985240d6fef/core/src/hlist.rs#L692-L716

Hirrolot:
Nice. Haven't seen this for the first time.

PR #168, opened 2020-08-08
Added a missing import in the Semigroup example

VitalyOlegovic:
The Semigroup example is missing the following import:
`use frunk::semigroup::All;`


lloydmeta:
Thanks !

PR #170, opened 2021-03-31
Allow folding hlist with a single `Poly`

ImmemorConsultrixContrarie:
Added a mention of `Poly` in docs, added implementations for both `foldl` and `foldr`, added simple tests.

Where it could be usable:

1. Folding a hlist of hlists with a single `Poly`.
2. Inversing a `map -> chain` iterator dependency (especially useful for by-value iterators mixed with by-ref iterators, where you don't want to clone referenced values):

```rust
use frunk_core::{
    traits::{Func, Poly},
    *,
};
use std::borrow::Borrow;

struct Sum;
impl<I, T> Func<(i32, I)> for Sum
where
    I: IntoIterator<Item = T>,
    T: Borrow<i32>,
{
    type Output = i32;

    fn call(args: (i32, I)) -> i32 {
        let (acc, iter) = args;
        iter.into_iter().fold(acc, |acc, x| acc + *x.borrow())
    }
}

fn main() {
    // Can do this:
    let h = hlist![0..10, [1, 2, 3, 4].iter()];
    let res = h.foldl(Poly(Sum), 0);

    // Can't do this:
    // let sum: i32 = (0..10).map(|x| x.borrow()).chain([1, 2, 3, 4].iter()).sum();
    //                                -^^^^^^^^^
    //                                |
    //                                returns a value referencing data owned by the current function
    //                                `x` is borrowed here
}
```

3. Probably more use cases I haven't thought of.


core/src/hlist.rs
```diff
@@ -441,6 +442,8 @@ macro_rules! gen_inherent_methods {
             /// acc
             /// ```
             ///
+            /// [`Poly`]: ../traits/struct.Poly.html
```
ExpHP, line 445: Perhaps at some point we should update these docs to strip these handwritten links, as rustc has supported intra-doc links since 1.48.

core/src/hlist.rs
```diff
@@ -1542,6 +1576,36 @@ mod tests {
...
     }
 
+    #[test]
+    fn test_poly_foldr_consuming() {
+        trait Dummy {
+            fn dummy(&self) -> i32 {
+                1
+            }
+        }
+        impl<T: ?Sized> Dummy for T {}
```
ExpHP, line 1586: Is this `?Sized` a leftover from some earlier version that made use of the bound?  (or is it just a "if you can, you might as well" sort of thing)

ExpHP (approved):
Only one very minor nit/question, otherwise lgtm.

core/src/hlist.rs
lloydmeta: Ah interesting...good point; https://blog.rust-lang.org/2020/11/19/Rust-1.48.html#easier-linking-in-rustdoc

```rust
pub mod foo {
    /// Some docs for `Foo`
    ///
    /// You may want to use `Foo` with [`Bar`](crate::bar::Bar).
    pub struct Foo;
}

pub mod bar {
    /// Some docs for `Bar`
    ///
    /// You may want to use `Bar` with [`crate::foo::Foo`].
    pub struct Bar;
}
```

lloydmeta (approved):
LGTM.

core/src/hlist.rs
ImmemorConsultrixContrarie: I copypasted this from `map` docs, no brain involved. Maybe you should open an issue about it or write PR to fix it if you really care.

core/src/hlist.rs
ImmemorConsultrixContrarie: It's "I have a habit of unconstraining generic trait impls whenever I write them"; mostly due to the first time I forgot to write `T: ?Sized` and got messy compiler errors.

core/src/hlist.rs
ExpHP: Yeah, no, it's just an off-hand comment, nothing specific to this PR at all

ExpHP:
*poke*

I just randomly remembered this exists out of the blue.  Do we want to merge?

lloydmeta:
Yep we do want to merge this; I've just kept it open for now because I don't have time yet to merge and release. 🙇‍♂️

PR #178, opened 2021-04-21
Make `foldl` and `foldr` take exactly same arguments

### LINKED ISSUE #171: Accumulator and head value generics are swapped for hlist `foldl` and `foldr`
It was especially irksome when I was writing #170, since you can't use the same generic poly for both folds:

```rust
struct Sum;
impl<Acc, T> Func<(Acc, T)> for Sum
where
    Acc: std::ops::Add<T, Output = Acc>
{
    type Output = Acc;

    fn call(args: (Acc, T)) -> Acc {
        args.0 + args.1
    }
}
```

Now you can only use this with `foldl`.
You can't even implement `Func<(T, Acc)> for Sum`, 'cuz generics;  a whole new `struct SumR` is needed to use `foldr`.
It is not so bad with a single closure, though you can't use the same closure for both and have to do argument reverse:

```rust
let rfolder = |t, init| lfolder(init, t);
```

So, was this a deliberate choice? Or would you accept a PR that reverses `T, Init` into `Init, T` for all right folds?

ImmemorConsultrixContrarie:
Closes #171.

Unresolved question: do we want [dirty hacks](https://github.com/lloydmeta/frunk/issues/171#issuecomment-812454214) to completely unify `foldl` and `foldr` arguments?

### DIFF
--- a/core/src/hlist.rs
+++ b/core/src/hlist.rs
@@ -1120,21 +1120,21 @@ where
 
 impl<F, R, H, Tail, Init> HFoldRightable<F, Init> for HCons<H, Tail>
 where
-    Tail: fn_foldr::FnHFoldRightable<F, Init>,
+    Tail: foldr_owned::HFoldRightableOwned<F, Init>,
     F: Fn(<Tail as HFoldRightable<F, Init>>::Output, H) -> R,
 {
     type Output = R;
 
     fn foldr(self, folder: F, init: Init) -> Self::Output {
-        fn_foldr::FnHFoldRightable::real_foldr(self, folder, init).0
+        foldr_owned::HFoldRightableOwned::real_foldr(self, folder, init).0
     }
 }
 
-/// [`HFoldRightable`] inner mechanics for folding with a single closure/function.
-pub mod fn_foldr {
+/// [`HFoldRightable`] inner mechanics for folding with a folder that needs to be owned.
+pub mod foldr_owned {
     use super::{HCons, HFoldRightable, HNil};
 
-    /// A real `foldr` for the folder that is a closure or a function.
+    /// A real `foldr` for the folder that must be owned to fold.
     ///
     /// Due to `HList` being a recursive struct and not linear array,
     /// the only way to fold it is recursive.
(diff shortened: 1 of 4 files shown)


ExpHP:
Regarding the "dirty hacks," yes, I think achieving parity between the two would be nice.

That said, I think I would prefer not seeing `#[doc(hidden)]` used to hide traits.  Frunk has always been pretty open with its traits, even when they are things we might prefer to be implementation details.  Seeing a trait bound in the docs that you can't click on can be kind of frustrating when there's no clue what it is for.  So I would suggest adding a note to it explaining precisely what functionality it enables (IIUC, Foldr with a single unborrowed `Fn`s) and discouraging users from using it directly.

core/src/hlist.rs
```diff
@@ -1108,40 +1108,81 @@ impl<F, FolderHeadR, FolderTail, H, Tail, Init> HFoldRightable<HCons<F, FolderTa
...
+    /// the `HList`:
+    ///
+    /// 1. `foldl` calls `folder(head)` and then passes the ownership
+    ///     of the folder to the next recursive call.
+    /// 2. `foldr` passes the ownership of the folder to the next recursive call,
+    ///     and then tries to call `folder(head)`; but the ownership is already gone!
+    ///
+    /// Thus, it should either take a reference to the folder or return ownership back.
+    /// This trait does the latter to avoid lifetimes in the trait definition.
+    pub trait FnHFoldRightable<Folder, Init>: HFoldRightable<Folder, Init> {
```
ImmemorConsultrixContrarie, line 1152: One more unresolved question.
This could be used later for other right-folds with ownership problems. The name and docs are closure/function related, though, and due to module and trait being public, changing names later would be a breaking change.
So, yeah, the question is: any ideas for better generic name and docs for this thing?

lloydmeta: This is a good point... I think it's generally good to use a name that is descriptive wrt the purpose of the trait and easily understood. Just going to throw this out there: "ClosureUnifier"?? No worries if it's horrible, just wanted to get the ball rolling :)

ExpHP: It's a tricky question.  HFoldRightRecoverable, since you recover the function?  I dunno.

Honestly though there already a lot of things frunk can't add without a semver breaking change, so it doesn't feel like a big deal to me. It's a wonder that frunk is only at 0.3, but iirc that's because it historically did not follow semver.

ImmemorConsultrixContrarie: Changed `fn_foldr` to `foldr_owned` and `FnHFoldRightable` to `HFoldRightableOwned`. Maybe not the clearest name, but it does decent job at describing the trait and is generic enough. If everyone's okay with the name, this conversation could be resolved, and the PR could be merged, I think.

ImmemorConsultrixContrarie:
By the way, since this PR is highly likely to get merged and is a breaking change, you may want to add a bunch of other [breaking changes](https://github.com/lloydmeta/frunk/issues?q=is%3Aissue+is%3Aopen+label%3Abreaking-change) to the patch until you push the new version, @lloydmeta.

lloydmeta (approved):
LGTM.

As mentioned, we should see which of [these](https://github.com/lloydmeta/frunk/issues?q=is%3Aissue+is%3Aopen+label%3Abreaking-change) are low hanging fruits to release w/ 0.4 (no major bump because we aren't at 1.0 yet, so breaking changes are fine with minor bmps)

lloydmeta:
There was a check failing due to a new clippy recommendation, which I fixed on `master`, merging into this branch to see if it passes..

@ImmemorConsultrixContrarie do you want to add a short for this change in https://github.com/lloydmeta/frunk/blob/master/CHANGELOG.md ?

PR #181, opened 2021-05-20
Implement folds for references

### LINKED ISSUE #180: Folds are not implemented for `&hlist` and `&mut hlist`
### Reasons to have it

The main reason to have it is performance.

1. Destructing `HCons<H, T>` into `H` and `T` copies both `H` and `T`.
2. Destructing `&HCons<H, T>` into `&H` and `&T` involves only pointer arithmetic with constants and one pointer copy.

```rust
hlist.to_ref().fold() // create a new hlist with a bunch of pointers, then destruct it.
(&hlist).fold() // pointer arithmetic
```

The compiler *might* optimize both lines into the same binary code, or it might not.

### Problems

The biggest problem will be UX. Right now `HCons` and `HNil` have nice methods `foldl` and `foldr`, and to use those methods the user doesn't have to import any traits.
With three fold impls for `hlist`, `&hlist` and `&mut hlist`, there should be either no method on the struct itself, or three methods with different names, kinda like `into_iter()`, `iter()` and `iter_mut()`.

Also, there will be some inconsistency between folds and other things like `HMappable`, which will be implemented for by-value hlists only; though `HMappable` could be implemented with `foldr`, and the usefulness of this trait is a bit questionable. `IntoReverse` could be implemented with `foldl`. `ToRef` and `ToMut` could be implemented with `foldr`.

### Problems of not having those impls

Some libraries may use hlist instead of tuples with

```rust
impl_trait_for_tuples(A, B, C, /* … */);
```

to make compilation time a bit better, or to avoid some unnecessary `unsafe` in those impls (because hlist could be folded, while tuple could not be).

However, generic bounds without fold impls for `&hlist` and `&mut hlist` could be messy (or won't compile at all without some serious thinking):

```rust
UserHlist: HFoldLeftable<ByValFolder, ValAcc> + for <'a> ToRef<'a>,
for <'a> <UserHlist as ToRef<'a>>::Output: HFoldLeftable<ByRefFolder, RefAcc>,
```

Then, the library will write its own `HFoldLeftable` (or `HFoldRightable`) due to either performance or generic problems. However, if every such library would write its own folding trait, the user who will try to use both of those libraries in the same function will have to write a bit of boilerplate:

```rust
fn foo<H>(h: &H)
where &H: lib1::HFoldLeftable<Folder, Acc> + lib2::HFoldLeftable<Folder, Acc>
{}
```

The traits are identical, but the user should write both of them in the bound \*sad user noises\*.

So, yeah, any thoughts about it, @lloydmeta, @ExpHP ?

ImmemorConsultrixContrarie:
Closes #180.

Things to do before this draft PR could be made an actual PR:

- [x] Implementation
- [ ] Tests
- [ ] Bikeshedding inherent methods. Do we want to completely remove those and force user to import fold traits (maybe like `use frunk::prelude::*;`)? Do we want to have something like `foldl_by_ref`, `foldl_by_mut`, `foldl_by_val`?
- [ ] Adjust documentation according to what was decided in the bikeshedding part

### DIFF
--- a/core/src/hlist.rs
+++ b/core/src/hlist.rs
@@ -1313,6 +1313,137 @@ where
     }
 }
 
+mod folds_for_ref {
+    use super::{
+        foldr_owned::{self, HFoldRightableOwned},
+        Func, HCons, HFoldLeftable, HFoldRightable, HNil, Poly,
+    };
+
+    macro_rules! ref_fold_impls {
+        ($( $mut_:tt )?) => {
+            impl<F, Acc> HFoldLeftable<F, Acc> for & $( $mut_ )? HNil {
+                type Output = Acc;
+
+                fn foldl(self, _: F, acc: Acc) -> Self::Output {
+                    acc
+                }
+            }
+
+            impl<'a, F, R, FTail, H, Tail, Acc> HFoldLeftable<HCons<F, FTail>, Acc> for &'a $( $mut_ )? HCons<H, Tail>
+            where
+                &'a $( $mut_ )? Tail: HFoldLeftable<FTail, R>,
+                F: FnOnce(Acc, &'a $( $mut_ )? H) -> R,
+            {
+                type Output = <&'a $( $mut_ )? Tail as HFoldLeftable<FTail, R>>::Output;
+
+                fn foldl(self, folder: HCons<F, FTail>, acc: Acc) -> Self::Output {
+                    let HCons { head, tail } = self;
+                    tail.foldl(folder.tail, (folder.head)(acc, head))
+                }
+            }
+
+            impl<'a, P, R, H, Tail, Acc> HFoldLeftable<Poly<P>, Acc> for &'a $( $mut_ )? HCons<H, Tail>
+            where
+                &'a $( $mut_ )? Tail: HFoldLeftable<Poly<P>, R>,
+                P: Func<(Acc, &'a $( $mut_ )? H), Output = R>,
+            {
+                type Output = <&'a $( $mut_ )? Tail as HFoldLeftable<Poly<P>, R>>::Output;
+
+                fn foldl(self, poly: Poly<P>, acc: Acc) -> Self::Output {
+                    let HCons { head, tail } = self;
+                    let r = P::call((acc, head));
+                    tail.foldl(poly, r)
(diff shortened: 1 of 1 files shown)


ImmemorConsultrixContrarie:
Funny fact: inherent `foldl` method does not interfere with `(&hlist).foldl` or `(&mut hlist).foldl` calls.
However, I still think that we should remove or rename the inherent methods; due to the same names for methods in struct and in trait, compiler could emit some [misleading errors](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=cfe7564b4586540997a579e04c643fb1)

lloydmeta:
Since there's no conflict, I would prefer that we keep this PR to just implementing the fold methods for references. 

IMO, if the user is using the "wrong" folder "function" with a given binding (e.g. hlist value but folder expects references), then they should get an error, which is the same as if you pass a reference-expecting closure when `.map`ing over a `vec.into_iter()`: there is no separate `map_val` or `map_mut`.

ExpHP:
> However, I still think that we should remove or rename the inherent methods; due to the same names for methods in struct and in trait, compiler could emit some misleading errors

I'm not at all keen on removing them, as I feel this can hurt discoverability.  Having them there [gives us this](https://docs.rs/frunk/0.3.2/frunk/hlist/struct.HCons.html); the whole API of HLists documented on a single page, in precisely the form that we are used to consuming documentation in as rustaceans.  Prior to 2.0, the closest you could get was to look at examples in README.md or on the hlist module&mdash;or to look at the traits in the docs, which were *scary.*

I would go so far as to say I'd rather have three inherent methods (`foldl`, `foldl_ref`, `foldl_mut`), and if I were to rename anything I'd rename the trait method to either `foldl_` or `foldl_generic`.  But this could be a slippery slope; there are other methods that could use `_ref` and `_mut` variants, yet obviously it'd be crazy to add these variants for *all* methods.  (probably just maps and folds?).  But personally this is why I've been satisfied with just having orthogonal `to_ref()/to_mut()` methods.

redbaron:
Been hit by the lack of folding on reference. I know it's been a while, but maybe there is a way forward to merge it? :)

PR #183, opened 2021-06-17
Bump quote, syn and proc-macro2 to 1

mbrobbel:
Bumps `quote`, `syn` and `proc-macro2` dependencies of Frunk's proc macro crates to `1`.

I'm working on a procedural macro where I want to use some public functions from the  Frunk procedural macro crates, but the function argument types are incompatible with version 1 of the `quote`, `syn` and `proc-macro2` dependencies.
Updating these dependencies is a breaking change for the users of these crates.

Thanks for this awesome crate!

### DIFF
--- a/proc-macro-helpers/src/lib.rs
+++ b/proc-macro-helpers/src/lib.rs
@@ -16,9 +16,8 @@ extern crate quote;
 extern crate syn;
 
 use proc_macro::TokenStream;
-use proc_macro2::TokenStream as TokenStream2;
+use proc_macro2::{Span, TokenStream as TokenStream2};
 use quote::ToTokens;
-use quote::__rt::Span;
 use syn::spanned::Spanned;
 use syn::{
     DeriveInput, Expr, Field, Fields, GenericParam, Generics, Ident, Lifetime, LifetimeDef, Member,
--- a/derives/Cargo.toml
+++ b/derives/Cargo.toml
@@ -15,8 +15,8 @@ travis-ci = { repository = "lloydmeta/frunk" }
 proc-macro = true
 
 [dependencies]
-syn = "0.15"
-quote = "0.6"
+syn = "1"
+quote = "1"
 
 [dependencies.frunk_proc_macro_helpers]
 path = "../proc-macro-helpers"
--- a/proc-macro-helpers/Cargo.toml
+++ b/proc-macro-helpers/Cargo.toml
@@ -12,9 +12,9 @@ keywords = ["Frunk", "macros", "internal"]
 travis-ci = { repository = "lloydmeta/frunk" }
 
 [dependencies]
-syn = "0.15"
-quote = "0.6"
-proc-macro2 = "0.4"
+syn = "1"
+quote = "1"
+proc-macro2 = "1"
 
 [dependencies.frunk_core]
 path = "../core"
--- a/proc-macros-impl/Cargo.toml
+++ b/proc-macros-impl/Cargo.toml
@@ -12,8 +12,8 @@ keywords = ["Frunk", "macros", "internal"]
 travis-ci = { repository = "lloydmeta/frunk" }
 
 [dependencies]
-syn = "0.15"
-quote = "0.6"
+syn = "1"
+quote = "1"
 proc-macro-hack = "0.5"
 
 [lib]


mbrobbel:
Looks like there were some updates to the [wrong_self_convention](https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention) lint in the [Clippy release for Rust 1.53](https://github.com/rust-lang/rust-clippy/blob/master/CHANGELOG.md#rust-153). Fixing those would be a breaking change.

ExpHP:
Certainly, quote, syn, and proc-macro2 have been v1.0 for quite a long time now, and it'd be nice for frunk to catch up.

> I'm working on a procedural macro where I want to use some public functions from the Frunk procedural macro crates, but the function argument types are incompatible with version 1 of the quote, syn and proc-macro2 dependencies.

This bit does bother me somewhat; I assume these functions are located in `frunk-proc-macro-helpers`, and I have to wonder if they were really written with any expectation that they might be used by code outside of frunk.  I wonder if some of these should be `#[doc(hidden)]`?

Which functions you are using?  Perhaps there is a cleaner alternative, or at least we can try to support some vague notion of stability on a subset of functions in the crate that are useful to other people.

mbrobbel:
> This bit does bother me somewhat; I assume these functions are located in frunk-proc-macro-helpers, and I have to wonder if they were really written with any expectation that they might be used by code outside of frunk. I wonder if some of these should be #[doc(hidden)]?

Yes one function that I'm using is located there and I'll push a commit in a minute that moves the derive methods to the helper crate so that I can also use them.
I agree that typically these functions should be non-public, but for my use case I think it makes sense.

> Which functions you are using? Perhaps there is a cleaner alternative, or at least we can try to support some vague notion of stability on a subset of functions in the crate that are useful to other people.

- `frunk_proc_macro_helpers::build_label_type`: I'm generating heterogeneous lists with fields (based on struct fields). I'm using this method to generate the label types for the field names.
- `impl_generic`: the trait that I'm deriving requires the type to implement `Generic`, but I prefer to not set this as a trait bound, because that makes it part of the public API. Instead I want to derive `Generic` for the type in the procedural macro using the code from the `frunk_derives` crate (by moving those impl functions to the helper crate).

ExpHP:
Rather than making the `impl_generic` functions public, is it not possible to have your proc macro emit `#[derive(Generic)]`?

mbrobbel:
> Rather than making the `impl_generic` functions public, is it not possible to have your proc macro emit `#[derive(Generic)]`?

Not that I'm aware of. You can only append to the input token stream in derive macros.

ExpHP:
I see, so yours is a derive macro.  Indeed, you cannot add a derive, then.

Thinking about this at first, I wondered: "Huh, this feels like a really poor limitation of derive macros!"  But then something occurred to me:  Imagine if there was another crate with a derive macro that had the same idea as yours, automatically adding a Generic impl for a type.  What would happen if a crate downstream wanted to derive both your trait and that trait?

```rust
use crate_1::Trait1;
use crate_2::Trait2;

// this could generate two conflicting Generic impls!
#[derive(Trait1, Trait2)]
struct MyType(i32, i32);
```

So actually, I feel like it is by design that you should have to treat Generic as a public dependency, and the user should be required to write `#[derive(Generic, Trait1, Trait2)]`.

mbrobbel:
That's a great point. You can probably code around it, but adding a trait bound seems like the correct solution.

proc-macro-helpers/src/lib.rs
```diff
@@ -16,9 +16,8 @@ extern crate quote;
 extern crate syn;
 
 use proc_macro::TokenStream;
-use proc_macro2::TokenStream as TokenStream2;
+use proc_macro2::{Span, TokenStream as TokenStream2};
 use quote::ToTokens;
-use quote::__rt::Span;
```
ExpHP, line 21: Oh wow, I can't believe we were importing from here without even a comment to justify why! 😬

mbrobbel:
> (one note: This does make me wonder now about thiserror's automatic Display impl, and whether this has the same design flaw? Perhaps that one is not so much of an issue because you have to opt into it with #[error] attributes...)

This triggers
```
conflicting implementations of trait `std::fmt::Display` for type `Foo`
```
when you also implement the Display trait for your type. (Display can't be derived, but the error would be the same)

proc-macro-helpers/src/lib.rs
lloydmeta: +1 Thanks for fixing this!

lloydmeta (approved):
LGTM.

lloydmeta:
Just an FYI, I plan on cutting a new release this Friday; sorry, just haven't had the time >.<

PR #186, opened 2021-06-18
Fix unicode identifiers support

### LINKED ISSUE #184: Cl failures from clippy
https://github.com/lloydmeta/frunk/runs/2853440952

Clippy is mad because HNil implements Copy, so an inherent method `HNil::to_ref(&self) -> Self` implies a costlier operation than it actually is.  What it doesn't understand is that this method is being generated by a macro, and needs to match similar methods on other types.

Probably just want to ignore this.

### LINKED ISSUE #185: Unicode identifiers
Rust 1.53 has finally [stabilized the `non_ascii_idents` feature](https://github.com/rust-lang/rust/pull/83799)!

I do recall that frunk's type-system encoding of strings already supports unicode.  Thus it'd be interesting to look into: Does frunk already support these identifiers?  If so, we should add some regression tests.  If not, we could look into supporting them.

lloydmeta:
* Fixes #185  (unicode support)
* Fixes #184 (clippy warning)
  * Probably unblocks the other PR.

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -121,7 +121,7 @@ pub struct InternalUser<'a> {
     pub age: usize,
     pub address: InternalAddress<'a>,
     pub is_banned: bool,
-    pub β: bool,
+    pub あ: bool,
 }
 
 #[derive(LabelledGeneric, PartialEq, Debug)]
@@ -140,7 +140,7 @@ pub struct ExternalUser<'a> {
     pub age: usize,
     pub address: ExternalAddress<'a>,
     pub name: &'a str,
-    pub β: bool,
+    pub あ: bool,
 }
 
 #[derive(LabelledGeneric, PartialEq, Debug)]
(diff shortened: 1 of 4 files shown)


tests/common/mod.rs
```diff
@@ -121,6 +121,7 @@ pub struct InternalUser<'a> {
     pub age: usize,
     pub address: InternalAddress<'a>,
     pub is_banned: bool,
+    pub α: bool,
```
ExpHP, line 124: hmm, maybe try β which doesn't seem to be in the list of confusables, to avoid these warnings

ExpHP (changes requested):

tests/common/mod.rs
lloydmeta: Good idea.

ExpHP:
Hmmm, that unicode "mixed script confusibles" warning is really pesky. :/

Looking at this: https://www.unicode.org/Public/security/13.0.0/confusablesSummary.txt.  I guess β (beta) is no good either because it looks like ß?  I'm hard pressed to find any character that is not on here...

Maybe we should just `#![allow(mixed_script_confusables)]` on the file.  It is just a test file after all...

lloydmeta:
Looks like あ did the trick.

Issue #198, opened 2022-08-22
Minimal superset of two Coproducts?


rosehuds:
In my library, I have a generic function that creates a superset of two coproducts, which come from its type arguments. Currently I'm using two CoproductEmbedder bounds, but since there are infinite coproduct types that satisfy any set of Embedder bounds, type inference fails, and users of the library have to specify the superset type themselves. This is bad on its own, but the function has 22 generic arguments, so the user ends up with quite a few underscores.

So, how can I get the compiler to infer the minimal superset (I suppose the union?) of some coproducts? I've been trying to make a new CoproductJoiner trait to help, but so far I've failed, and I suppose the maintainers here are likely better at this sort of thing than I am :D

ExpHP:
If you're only ever using this with completely disjoint sets of types, this should be possible.

However, I presume you included the word "minimal" for a reason...  and unfortunately, any attempt to join two coproducts without duplicates smells impossible.  By which I mean, it sounds similar to numerous other things that have previously been found to be impossible to implement in frunk due to limitations of rust's trait system.

In particular:

* If you want `CoproductJoiner<Coprod![B]> for Coprod![A]` to produce `Coprod![A, B]`,
* ...but also want `CoproductJoiner<Coprod![A]> for Coprod![A]` to produce `Coprod![A]`,

then almost certainly, any attempt to make the first bullet work for *most* `B` will have parametricity over `B`, causing it to apply to *all* `B`, including `B=A`; which will conflict with the second bullet.  Frunk's use of index types to disambiguate impls does allow it to overcome some problems of overlapping impls, but even this technique cannot be used to break parametricity.

---

> 22 generic arguments

Ahh-e-uh.  *Yikes!*

I don't really have a nice alternative to recommend here.

Depending on your circumstances, *perhaps* one possible compromise is to use a single superset coproduct type throughout the higher-level parts of the application, to reduce the number of places where distinct type annotations must be maintained.  (however, this also forfeits many of the benefits of Coproduct, and at some point one must ask themselves if they'd be better off with a regular `enum`).

[...]

PR #199, opened 2022-09-07
Optimized memory layout for Coproducts

joonazan:
Instead of nesting enums, the new representation explicitly has one tag and nested unions. The nested unions only take up as much space as the largest variant of the coproduct, unlike the old representation, which was optimized very poorly by the compiler.

For example the below line takes 120 bytes on frunk 0.4 but only 16 bytes with my implementation. If I change the tag from `usize` to `u32` it drops to 8 bytes. This raises the question whether it is possible to dynamically choose a good tag size. It is probably not worth the complexity, though, as I suspect that tags larger than `u16` aren't practical simply because the data structure's size would slow down compilation too much.

```Rust
Coprod! {u8, u8, u8, u8, u32, u8, u8, u8, u8, u8, u8, u8, u8, u8, u8, u8, u32, u8, u8, u8, u8, u8, u8, u8, u8, u8, u8, u8, u32, u8, u8, u8, u8, u8, u8, u8};
```

The code passes all tests but it is still quite messy and some of the old Coproduct's API like Ord and Eq is still missing.

I also need to name the current Coproduct LeakingCoproduct and put it in yet another wrapper that has a custom Drop implementation. That, or make Coproduct only take values that are Copy.

### DIFF
--- a/core/src/coproduct.rs
+++ b/core/src/coproduct.rs
@@ -71,9 +71,132 @@
 //! # }
 //! ```
 
+use core::mem::ManuallyDrop;
+
 use crate::hlist::{HCons, HNil};
 use crate::indices::{Here, There};
 use crate::traits::{Func, Poly, ToMut, ToRef};
+
+/// Coproduct that does not know what variant it currently holds.
+/// Note: contents must be manually dropped!
+
+// C representation is required to ensure that the same variant is represented
+// in exactly the same way in all coproducts.
+// See https://rust-lang.github.io/unsafe-code-guidelines/layout/unions.html
+#[repr(C)]
+pub union UntaggedCoproduct<H, T> {
+    head: ManuallyDrop<H>,
+    tail: ManuallyDrop<T>,
+}
+
+/// Phantom type for signature purposes only (has no value)
+///
+/// Used by the macro to terminate the Coproduct type signature
+#[derive(PartialEq, Debug, Eq, Clone, Copy, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum CNil {}
+
+pub trait IndexedDrop {
+    fn idrop(&mut self, i: usize);
+}
+
+impl IndexedDrop for CNil {
+    fn idrop(&mut self, _: usize) {}
+}
+
+impl<H, T: IndexedDrop> IndexedDrop for UntaggedCoproduct<H, T> {
+    fn idrop(&mut self, i: usize) {
+        if i == 0 {
+            unsafe { ManuallyDrop::drop(&mut self.head) }
+        } else {
+            unsafe { &mut self.tail }.idrop(i - 1)
+        }
+    }
+}
+
+pub trait Injector<InjectType, Index> {
+    fn inject(to_insert: InjectType) -> Self;
+}
+
+impl<I, Tail: IndexedDrop> Injector<I, Here> for UntaggedCoproduct<I, Tail> {
+    fn inject(to_insert: I) -> Self {
+        UntaggedCoproduct {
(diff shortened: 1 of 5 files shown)


joonazan:
I think I understand the design space of this change relatively well now.

Instead of Coproduct, there needs to be Coproduct and CopyableCoproduct, which are otherwise identical but Coproduct is Drop and CopyableCoproduct is Copy.

The trait bounds may have to get uglier (as seen in my current code). The problem is that I have traits implemented on UntaggedCoproduct and I can not figure out how to make `Coproduct<Untagged>: Trait2` imply `Untagged: Trait1`.

Some unsafe is necessary because that is the only way to use unions.

Do you think this is good for frunk, or should I rather make this a separate coproduct library?

ExpHP:
Hmmm.  This is a tough one.

@lloydmeta, would it be correct to say that "no usage of `unsafe`" is one of the design goals of Frunk?

I've always seen frunk as the kind of library that shows what's possible with type-level programming in rust, even if it is not necessarily the best implementation.

All of this said, there's no mistaking that the current size of Coproducts is... well, not good.   And I don't like to think that our advice to people can only be "don't use frunk's `Coproduct`."  Improving the size of Coproduct could also improve the performance of Generic for people who are using that.  These could potentially serve as arguments for making such a change to frunk.

joonazan:
The README implies that no unsafe is a goal of frunk. That most likely loses performance elsewhere as well, so using unsafe and having a worse API just to improve performance seems like a bad fit for frunk. The only counterargument I can think of is that the huge proc macros generated by frunk can be seen as somewhat dangerous, though not in the same way as unsafe.

Anyway, I'm glad that frunk exists because otherwise I wouldn't have figured out that the data structure implemented in this PR is possible in Rust. It is good to have an easy to understand implementation that shows what is possible in Rust. My version of coproducts is not as easy to understand, as it uses the same construction as the safe version on the type level but the actual implementation has to be wildly unsafe.

My use case is that I've written [a subset of an algebraic effect library](https://github.com/joonazan/alternator) (stable and no unsafe btw) and seen [a more fully featured one](https://github.com/rosefromthedead/effing-mad) that uses frunk's Coproduct. I want to try using algebraic effects really heavily, but that requires that they are very cheap.

ExpHP:
Interesting.  Yeah an effect system library definitely sounds like the kind of thing that needs this.  Funny, though, I would've expected them to require some concept of "minimal common superset" (which unfortunately is [also a huge technical challenge in rust](https://github.com/lloydmeta/frunk/issues/198)).

Anyways, best of luck with that!

joonazan:
I made the [standalone version](https://docs.rs/coproduct/). Still missing some features; I'll add them as needed. I think this PR can be closed now.

I will also make a PR for using my library in effing-mad sometime soon. I haven't used my own library much yet, so that will hopefully be a better test.

@ExpHP I looked into computing the union of two coproducts. It is currently impossible but most of it is pretty straightforward. The impossible part is behaving differently when two types are equal vs when they are not.

Type comparison is easy via the specialization feature, which is unfortunately currently unsound. I think this use case would be sound as it doesn't involve lifetimes but who knows.

Alternatively one could assign compile-time natural numbers to all types that go into the coproduct.

ExpHP:
> Type comparison is easy via the specialization feature, which is unfortunately currently unsound. I think this use case would be sound as it doesn't involve lifetimes but who knows.

As far as I am aware, the current plans to implement a sound subset of specialization explicitly forbid any attempt to specialize on type equality, [because this can be used to equate lifetimes](http://smallcultfollowing.com/babysteps/blog/2018/02/09/maximally-minimal-specialization-always-applicable-impls/#condition-3-each-type-parameter-can-only-be-used-once).

ExpHP:
> I think this PR can be closed now.

Sure.

PR #201, opened 2022-09-22
add `extract` for getting the value out of a 1-coproduct

rosefromthedead:
I would find it useful :) It's like `.take().unwrap()` but without the implication that it might panic.

### DIFF
--- a/core/src/coproduct.rs
+++ b/core/src/coproduct.rs
@@ -581,6 +581,30 @@ impl<Head, Tail> Coproduct<Head, Tail> {
     }
 }
 
+impl<T> Coproduct<T, CNil> {
+    /// Extract the value from a coproduct with only one variant.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # fn main() {
+    /// use frunk_core::Coprod;
+    ///
+    /// type I32Only = Coprod!(i32);
+    /// let co = I32Only::inject(5);
+    ///
+    /// assert_eq!(co.extract(), 5);
+    /// # }
+    /// ```
+    #[inline(always)]
+    pub fn extract(self) -> T {
+        match self {
+            Coproduct::Inl(v) => v,
+            Coproduct::Inr(never) => match never {},
+        }
+    }
+}
+
 /// Trait for instantiating a coproduct from an element
 ///
 /// This trait is part of the implementation of the inherent static method


lloydmeta:
Interesting; can you please explain your use case for making a Coproduction of 1 type?

ExpHP:
Idk this is the author of [effing-mad](https://github.com/rosefromthedead/effing-mad) so if they have a use case, I believe them 🤣

rosefromthedead:
Sorry, I tried to reply this morning, but my email server fell over :)

My API involves generic coproducts, and users* provide functions from a subset of a coproduct to a subset of another - providing multiple functions until all variants are accounted for. If the user wants to deal with only one variant of each, it is more ergonomic for me to provide API surface that does the extracting and injecting around such a function, so the user doesn't have to. These APIs are internally implemented using the more general public API, so my code has a fair few match expressions that do what `extract` could do.

*users (hypothetical). No one actually uses it yet :)

core/src/coproduct.rs
```diff
@@ -581,6 +581,30 @@ impl<Head, Tail> Coproduct<Head, Tail> {
...
+    /// let co = I32Only::inject(5);
+    ///
+    /// assert_eq!(co.extract(), 5);
+    /// # }
+    /// ```
+    #[inline(always)]
+    pub fn extract(self) -> T {
+        match self {
+            Coproduct::Inl(v) => v,
+            Coproduct::Inr(never) => match never {},
```
lloydmeta, line 603: Not really sure which is better, but I've always reached for [`unreachable!`](https://doc.rust-lang.org/std/macro.unreachable.html) for this sort of thing?

```suggestion
            Coproduct::Inr(never) => unreachable!(),
```

lloydmeta (approved):
LGTM, and thanks for clarifying your use case (it looks ultra cool btw), and sorry for my late/delayed response(s); life has been kicking my butt lately.

I'm going to merge this in the next day or so, and add a reminder for myself to push out a new release in at most a month (again, sorry, there's a lot going on).

core/src/coproduct.rs
ExpHP: Hmm, personally I disagree.  `match never {}` *proves* that something is unreachable, whereas it is possible for `unreachable!()` to be used in reachable code.

rosefromthedead:
Yeah, I used to use `unreachable!()` for these cases, until someone sent me a patch to do it this way. I agree that since `match never {}` can only be used when it is definitely impossible, it better conveys the semantics of that match arm.
-------- Original message --------From: Michael Lamparski ***@***.***> Date: 01/10/2022  18:39  (GMT+00:00) To: lloydmeta/frunk ***@***.***> Cc: Rose Hudson ***@***.***>, Author ***@***.***> Subject: Re: [lloydmeta/frunk] add `extract` for getting the value out of a 1-coproduct (PR #201) 
@ExpHP commented on this pull request.



In core/src/coproduct.rs:
> +    ///
+    /// ```
+    /// # fn main() {
+    /// use frunk_core::Coprod;
+    ///
+    /// type I32Only = Coprod!(i32);
+    /// let co = I32Only::inject(5);
+    ///
+    /// assert_eq!(co.extract(), 5);
+    /// # }
+    /// ```
+    #[inline(always)]
+    pub fn extract(self) -> T {
+        match self {
+            Coproduct::Inl(v) => v,
+            Coproduct::Inr(never) => match never {},

Hmm, personally I disagree.  match never {} proves that something is unreachable, whereas it is possible for unreachable!() to be used in reachable code.

core/src/coproduct.rs
lloydmeta: Thanks both; TIL! Makes total sense to me.

PR #204, opened 2022-11-03
Add `Coproduct::map()` method

### LINKED ISSUE #203: `Coproduct::map()`
I'm tinkering around with developing an algebraic effect system in Rust.

Heavy inspiration is taken from the old [`eff`](https://github.com/pandaman64/effective-rust) project. The basic idea is that each effectful function is a `Generator` that yields `Coprod!(Each, Different, Effect, Raised)`:

```rust
struct A;
struct B;
struct C;

impl Effect for A {
    type Output = ();
}
impl Effect for B {
    type Output = i32;
}
impl Effect for C {
    type Output = ();
}

fn make_effectful_function() -> impl Effectful<Output = String, Effect = Coprod!(A, B, C)> {
    || {
        let a: () = raise!(A);
        let b: i32 = raise!(B);
        let c: () = raise!(C);

        b.to_string()
    }
}
```

With each of those `raise!` macros expanding to something like:

```rust
{
    type Effect = Coprod!(A, B, C);
    type Output = Coprod!((), i32, ());

    let (send, recv): (Sender<Effect>, Receiver<Output>) = channel();
    yield (Effect::inject(B), send);
    let output = recv.recv();
    let output = output.uninject::<i32, _>(output); // in reality this remembers the Index used to inject B
    output
}
```

## `Coproduct::map()` for handling effects

Whoever drives these effectful computations needs to handle the raised effects, mapping each possible effect to its proper output type. Here's how that could work if `Coproduct` had a `map()` like `HList` does:

```rust
let mut effectful = make_effectful_function();

let result = loop {
    match effectful.poll() {
        Poll::Complete(result) => break result,
        Poll::Effect(effect, resume) => {
            let output = effect.map(poly![
                |A| (),
                |B| 100,
                |C| (),
            ]);
            resume.send(output);
        },
    }
}
```

# Way forward?

I'd be willing to take a stab at adding these bits of functionality, but don't know how I'd best go about that. Like, should I repurpose `HMappable` to also work with `Coproduct`? Or should I create a brand new `CMappable`?

BGR360:
Also, make `poly_fn!` accept more inputs along the way.

Closes #203.

### DIFF
--- a/core/src/coproduct.rs
+++ b/core/src/coproduct.rs
@@ -523,11 +523,11 @@ impl<Head, Tail> Coproduct<Head, Tail> {
     /// # fn main() {
     /// use frunk_core::{Coprod, hlist};
     ///
-    /// type I32F32StrBool = Coprod!(i32, f32, bool);
+    /// type I32F32Bool = Coprod!(i32, f32, bool);
     ///
-    /// let co1 = I32F32StrBool::inject(3);
-    /// let co2 = I32F32StrBool::inject(true);
-    /// let co3 = I32F32StrBool::inject(42f32);
+    /// let co1 = I32F32Bool::inject(3);
+    /// let co2 = I32F32Bool::inject(true);
+    /// let co3 = I32F32Bool::inject(42f32);
     ///
     /// let folder = hlist![|&i| format!("int {}", i),
     ///                     |&f| format!("float {}", f),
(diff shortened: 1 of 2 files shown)


core/src/coproduct.rs
```diff
@@ -523,11 +523,11 @@ impl<Head, Tail> Coproduct<Head, Tail> {
     /// # fn main() {
     /// use frunk_core::{Coprod, hlist};
     ///
-    /// type I32F32StrBool = Coprod!(i32, f32, bool);
+    /// type I32F32Bool = Coprod!(i32, f32, bool);
```
lloydmeta, line 526: 👍🏼

lloydmeta (approved):
LGTM.

PR #208, opened 2023-02-28
Implement applyl_at and applyr_at functions

redbaron:
It adds `applyl_at` and `applyr_at` functions which apply single argument function to the `n`th HList element. It is implemented as a fold function which carries target and current depth levels as well as function application result in the accumulator.

Because `n` can be outside of the hlist length, function returns `Option<_>` with `None` returned if `n` is larger than list size.

It will also be possible to create `apply*_at_args` version which adds `args` tuple to the accumulator to be passed to the function as a poor replacement for missing `self` (and therefore closure) param.

This is a draft PR to collect early feedback, more specifically:
- is it something you'd like to see in this crate?
- should it be standalone function like it is now or you prefer it to be a method on the `HCons`?

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -73,5 +73,7 @@ pub mod path;
 pub mod traits;
 mod tuples;
 
+pub mod utils;
+
 #[cfg(test)]
 mod test_structs;
--- a/core/src/utils.rs
+++ b/core/src/utils.rs
@@ -1,0 +1,82 @@
+use core::marker::PhantomData;
+
+use crate::{
+    hlist::{HFoldRightable, HList},
+    traits::{Func, Poly},
+};
+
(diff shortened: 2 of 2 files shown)


redbaron:
Added `apply*_args_at` variants

lloydmeta (commented):
Out of curiosity, what is the use case for this? 

In general, I'd like to avoid failable-at-runtime things in this crate because it feels like we could avoid them since we (should) know almost everything at compile time. For instance, we could use a Labelled Hlist, pluck out the element we want by type (we'd get a compile time error if it does not exist or is duplicated), then apply a function?

redbaron:
Not everyting is known at compile time. In my experiments I have a  "query plan" (list of operations) in the HList and need to execute them in order when right conditions arise, hence the runtime `n` param to pick one to run.

PR #209, opened 2023-05-04
add function for explicitly extending an hlist

CobaltCause:
This is basically [`std::iter::Extend`](https://doc.rust-lang.org/stable/std/iter/trait.Extend.html) but for HLists.

### DIFF
--- a/core/src/hlist.rs
+++ b/core/src/hlist.rs
@@ -621,6 +621,33 @@ macro_rules! gen_inherent_methods {
             {
                 HFoldRightable::foldr(self, folder, init)
             }
+
+            /// Extend the contents of this HList with another HList
+            ///
+            /// This exactly the same as the [`Add`][Add] impl.
+            ///
+            /// [Add]: struct.HCons.html#impl-Add%3CRHS%3E-for-HCons%3CH,+T%3E
+            ///
+            /// # Examples
+            ///
+            /// ```
+            /// use frunk_core::hlist;
+            ///
+            /// let first = hlist![0u8, 1u16];
+            /// let second = hlist![2u32, 3u64];
+            ///
+            /// assert_eq!(first.extend(second), hlist![0u8, 1u16, 2u32, 3u64]);
+            /// ```
+            pub fn extend<Other>(
+                self,
+                other: Other
+            ) -> <Self as Add<Other>>::Output
+            where
+                Self: Add<Other>,
+                Other: HList,
+            {
+                self + other
+            }
         }
     };
 }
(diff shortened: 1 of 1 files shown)


core/src/hlist.rs
```diff
@@ -1546,6 +1569,48 @@ where
...
+/// [`HCons::extend`]: struct.HCons.html#method.extend
+pub trait HExtend<T: HList> {
+    /// The result of the combination
+    type Output: HList;
+
+    /// Extends `self` with `other`
+    fn extend(self, other: T) -> Self::Output;
+}
+
+impl<T: HList> HExtend<T> for HNil {
```
lloydmeta, line 1586: Nice, can you please add a test case for HNil extend HNil and Hcons extend HNil cases?

lloydmeta (commented):
Looking good to me, and thanks for adding this ! 

My only suggestion is to add some test cases for the "trivial" cases (doesn't need to be in the form of a doctest)

core/src/hlist.rs
CobaltCause: Done, with bonus HNil extend HCons case too.

core/src/hlist.rs
```diff
@@ -621,6 +621,29 @@ macro_rules! gen_inherent_methods {
             {
                 HFoldRightable::foldr(self, folder, init)
             }
+
+            /// Extend the contents of this HList with another HList
```
lloydmeta, line 625: Sorry, it's been a while, so I forgot about this, but there's an implementation of `Add` that looks identical: can we re-implement this method in terms of that?

https://github.com/lloydmeta/frunk/blob/341df3e921fee3e5891724abf4906869fa18284f/core/src/hlist.rs#L762-L786

CobaltCause: Oh... I didn't even see this impl; it indeed does the exact same thing as the `Extend` trait. Do you still want an inherent `extend` method? It might help with the discoverability problem I just had lol. Or adding something to the `hlist` module docs about the `Add` impl might be enough too.

lloydmeta: > Oh... I didn't even see this impl; it indeed does the exact same thing as the `Extend` trait. Do you still want an inherent `extend` method? It might help with the discoverability problem I just had lol. Or adding something to the `hlist` module docs about the `Add` impl might be enough too.

I would love to keep the inherent `extend` method. It 100% helps with discoverability.

lloydmeta (approved):
LGTM, thanks !

PR #216, opened 2023-06-23
avoid clippy::unneeded_field_pattern

figsoda:
see https://github.com/rust-lang/rust-clippy/issues/11008

### DIFF
--- a/core/src/macros.rs
+++ b/core/src/macros.rs
@@ -70,7 +70,14 @@ macro_rules! hlist_pat {
     () => { $crate::hlist::HNil };
     (...) => { _ };
     (...$rest:pat) => { $rest };
+    (_) => { $crate::hlist_pat![_,] };
     ($a:pat) => { $crate::hlist_pat![$a,] };
+    (_, $($tok:tt)*) => {
+        $crate::hlist::HCons {
+            tail: $crate::hlist_pat![$($tok)*],
+            ..
+        }
+    };
     ($a:pat, $($tok:tt)*) => {
         $crate::hlist::HCons {
             head: $a,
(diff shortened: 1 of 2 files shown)


lloydmeta (commented):
Overall LTGM, can you please add a test to check that this fixes the problem?

figsoda:
I added a test, but I don't think it would run in CI, but I did find a bug with my fix from the test which I now have fixed

lloydmeta:
> I added a test .. but I did find a bug from the test which I now have fixed

Thanks ! Great that the test caught the bug!

> but I don't think it would run in CI

Why do you think so? 

https://github.com/lloydmeta/frunk/actions/runs/5352333868/jobs/9707151131?pr=216#step:5:185 seems to indicate your new test is running.

core/src/macros.rs
```diff
@@ -405,4 +412,10 @@ mod tests {
         },));
         assert_eq!(h2, hlist![true, false, false]);
     }
+
+    #[test]
+    fn unneeded_field_pattern() {
+        #[deny(clippy::unneeded_field_pattern)]
+        let hlist_pat![_, _] = hlist![1, 2];
```
lloydmeta, line 419: Just for completeness, can we have another test that is partial `_`, like the [issue](https://github.com/rust-lang/rust-clippy/issues/11008) @CobaltCause filed?

figsoda: added a test case similar to the issue

figsoda:
> Why do you think so?

The error only shows up when I run `cargo clippy --tests --all`, but it seems like `--tests` is not passed to `cargo clippy` in CI

lloydmeta:
Ah in that case I think you need to update this bit

https://github.com/lloydmeta/frunk/blob/df2d6ea6242c853ea16fca944b2b1ef2af939fdf/.github/workflows/ci.yml#L23-L27

core/src/macros.rs
```diff
@@ -405,4 +412,12 @@ mod tests {
         },));
         assert_eq!(h2, hlist![true, false, false]);
     }
+
+    #[test]
+    #[deny(clippy::unneeded_field_pattern)]
+    fn unneeded_field_pattern() {
+        let hlist_pat![_, _] = hlist![1, 2];
+        let hlist_pat![foo, _, baz] = hlist!["foo", "bar", "baz"];
+        let _ = (foo, baz);
```
lloydmeta, line 421: Is there any chance you could add an assertion to the values of the bindings? It would give us more confidence about the macro changes.

lloydmeta (approved):

Issue #233, opened 2024-05-05
Discussion: shall we make std an optional extension?


Ben-PH:
It seems to me that this project deserves to be a no-std crate by default. From what I can gather, only a few things rely on `alloc` (see also #220 ), with no reliance on the standard library, at least when it comes to the core-concepts being dealt with. 

In terms of just `HList`, the only non-core items I came across is the into-vec implementation, and that should probably be updated to `alloc`-gated per #220 

Is it worth being `no-std` first and foremost to maximize compatibility, adding `no-std` to the cargo-toml keyword list, etc.?

Issue #220, opened 2023-07-13
`Validated`/`no_std`: replace `std::Vec` dependency with `alloc::Vec`


mammothbane:
Would you be open to a PR to conditionally replace the use of `std::Vec` in `frunk::Validated` with `alloc::Vec`? At a quick glance, I believe this is the only required change to make the whole project work in `no_std`-but-`alloc` environments.

ExpHP:
Does it need to be conditional?  I'm not familiar with usage of or best practices involving the alloc crate, but I would suspect that if it is possible to always use alloc over std when possible then that would be preferred.

ExpHP:
Oh yeah, I just remembered we have this:

```rust
#[cfg(not(feature = "std"))]
extern crate core as std;
```

Not 100% sure but I think that's a relic from a time back when you couldn't use `core::` without `no_std`??  These days you can use `core::` from any crate and that's what I think we ought to be doing for the things that don't require `std`.

PR #236, opened 2024-07-30
rework #![no_std] support, switch on `alloc`

mammothbane:
Finally got around to change I brought up in #220. See also #233.

Most of the uses of `std` in the crate are of `Vec`, `String`, and `Box`, which are in `alloc`, which is available on `#![no_std]` via `extern crate alloc;`. This commit adjusts these uses to use `alloc` directly and adds an `alloc` feature flag to control them. The notable exceptions still requiring a `std` flag are `HashMap` and `HashSet` (I'm working on a separate change to add `BTree{Map,Set}` support for `#![no_std]` envs). `frunk_core` no longer needs a `std` flag: it has been marked deprecated with a comment.

The reimport of `core` as `std` (see #220 for historical explanation) is converted to direct usage of `core` everywhere.

Removed `#[cfg(feature = "std")]` condition from `frunk_laws` and made it depend on `frunk/std` -- it depends on `quickcheck`, which requires `std`.

Added `#[cfg(test)] extern crate std;` to `#![no_std]` crates, as `std` is required to run the libtest harness.

Tested against all the combinations of feature flags I could think of. Successfully compiles into a `#![no_std]` embedded project with `alloc` enabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "frunk"
 edition = "2021"
-version = "0.4.2"
+version = "0.4.3"
 authors = ["Lloyd <lloydmeta@gmail.com>"]
 description = "Frunk provides developers with a number of functional programming tools like HList, Coproduct, Generic, LabelledGeneric, Validated, Monoid, Semigroup and friends."
 license = "MIT"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,10 +41,15 @@ version = "0.5.0"
 serde = { version = "^1.0", optional = true, features = [ "derive" ] }
 
 [features]
-default = ["validated", "proc-macros"]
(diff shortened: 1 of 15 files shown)


lloydmeta:
Thanks for this; I'll review more once I get a chance, but IIUC this should probably result in at least a minor version bump right?

mammothbane:
Yeah, everything other than removing the `frunk_core/std` should be minor &mdash; everything's compatible. Removing `frunk_core/std` is major, but I can add it back and mark as deprecated/to be removed in next major version.

lloydmeta (approved):
LGTM. Thanks for this. I'll keep this open for now just to make sure I don't miss it; planning to cut a new release by end of month.

lloydmeta:
Thanks for this. Just cut a new release for this.

PR #240, opened 2025-03-01
Implemented a reference version of Plucker/ByNameFieldPlucker

kgtkr:
I want to use Plucker for `&HCons<..., HCons<..., HCons<...>>>`. How about something like this?

### DIFF
--- a/core/src/hlist.rs
+++ b/core/src/hlist.rs
@@ -951,7 +951,7 @@ where
             <&'a Tail as Plucker<&'a FromTail, TailIndex>>::Remainder,
         ) = <&'a Tail as Plucker<&'a FromTail, TailIndex>>::pluck(&self.tail);
         (
-            &target,
+            target,
             HCons {
                 head: &self.head,
                 tail: tail_remainder,
--- a/core/src/hlist.rs
+++ b/core/src/hlist.rs
@@ -929,11 +929,11 @@ where
 }
 
 /// Implementation when target is reference and  the pluck target is in head
-impl<'a, T, Tail> Plucker<&'a T, Here> for &'a HCons<T, Tail> {
-    type Remainder = &'a Tail;
+impl<'a, T, Tail: ToRef<'a>> Plucker<&'a T, Here> for &'a HCons<T, Tail> {
+    type Remainder = <Tail as ToRef<'a>>::Output;
 
     fn pluck(self) -> (&'a T, Self::Remainder) {
-        (&self.head, &self.tail)
+        (&self.head, self.tail.to_ref())
     }
 }
 
@@ -1611,6 +1611,7 @@ where
 mod tests {
     use super::*;
 
+    use alloc::string::ToString;
     use alloc::vec;
 
     #[test]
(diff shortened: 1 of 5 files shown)


tests/generic_tests.rs
```diff
@@ -85,3 +90,31 @@ fn test_mixed_conversions_round_trip() {
     let u_again: SavedUser = convert_from(au);
     assert_eq!(u_again, before)
 }
+
+#[test]
+fn test_pluck() {
```
kgtkr, line 95: I'm not sure if this is the right place to write a test. I added it as a usage example, but I'll delete it if it's not needed.

lloydmeta (approved):
Thanks for sending this. It's been a while since these were implemented so it could just be memory loss on my part but I'm a bit surprised that we didn't have this 😅 

On the whole this makes sense, I just want some time to double check the above.

tests/labelled_tests.rs
```diff
@@ -277,3 +278,30 @@ fn test_transmogrify_keyword_field_embedder_structs() {
...
+        first_name: "Humpty",
+        last_name: "Drumpty",
+        age: 3,
+    };
+    let h = &into_labelled_generic(u);
+    let first_name: (Field<(f, i, r, s, t, __, n, a, m, e), _>, _) = h.pluck_by_name();
+    assert_eq!(first_name.0.value, &"Humpty");
+    let age: (Field<(a, g, e), _>, _) = h.pluck_by_name();
+    assert_eq!(age.0.value, &3);
+}
```
lloydmeta, line 307: I think having it here in `tests` is great but it might be good to also have (unit?) testing without using the `into_labelled_generic` machinery like so

https://github.com/lloydmeta/frunk/blob/ffba8cfa144b8ed02942271a3ea45006100cd029/core/src/labelled.rs#L972-L977

core/src/labelled.rs
```diff
@@ -976,6 +1011,32 @@ mod tests {
         assert_eq!(name.value, "Joe")
     }
 
+    #[test]
+    fn test_pluck_by_name() {
```
kgtkr, line 1015: Tests have been moved

core/src/hlist.rs
```diff
@@ -928,6 +928,38 @@ where
     }
 }
 
+/// Implementation when target is reference and  the pluck target is in head
+impl<'a, T, Tail: ToRef<'a>> Plucker<&'a T, Here> for &'a HCons<T, Tail> {
+    type Remainder = <Tail as ToRef<'a>>::Output;
```
kgtkr, line 933: While writing the test, I noticed that `Remainder` returns a value that is not an hlist. (example: `HCons<&String, &HCons<&i32, &HNil>>`)
I fixed it by adding `ToRef` so that it returns an hlist. (example: `HCons<&String, HCons<&i32, HNil>>`)
The same goes for `ByNameFieldPlucker`.

lloydmeta (commented):
Can you please add a note to the changelog?

kgtkr:
I have fixed the ChangeLog.

lloydmeta:
おまてせ 🙇🏻‍♂️ Just released this as part of 0.4.4

PR #254, opened 2026-07-04
Fix typo in coproduct docs

AMDmi3:
100% human spotted while reading the docs and 100% human fixed by hand


lloydmeta (approved):
Love it.  Thanks.  I guess no one was reading these hand-written comments back in the day 🧓🏼 .

